PR #10, opened 2022-08-02
Updates to RequestView table and RecentDeliveries table

ao508:
General refactoring done:
- use of react table components
- organized content by pages

Added logos.

Added a navbar.

Misc:
- placeholder for home page content

Signed-off-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

--

TODO:
- [x] switch to functional components (remove classes)
- [ ] also type functions as React.FunctionComponent (see comment below)
- [x] helper functions should be defined outside of components 

**Moved to separate cards:**
- add offset/limit to [graphql query](https://neo4j.com/docs/graphql-manual/current/pagination/offset-based/) instead of in react table 
- use moment.js for date formatting **--> made into separate card**
- use of `types` instead of classes for extending data models (might not need to do this if what we need is achieved by other means) // **NOTE: most of these were switched to `type` declarations but there are still a couple cases where this hasn't been done yet because of uncertainty on how to convert one data model `type` to another**  **--> made into separate card**


src/pages/recentDeliveries/RecentDeliveriesPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { loading, error, data } = useQuery(ExampleQueryDocument);
+
+  function transformAndFilterRequestsByDate(requestsList: Array<Request>) {
+    const referenceDate = new Date();
+    referenceDate.setDate(referenceDate.getDate() - 60);
+    const refDateAsDate = new Date(referenceDate);
+    let month = refDateAsDate.getMonth() + 1;
+    let day = refDateAsDate.getDate();
+
+    const refDateFilter = `${refDateAsDate.getFullYear()}-${
```
alisman, line 43: I would use Moment.js for date formatting.  Save yourself a lot of trouble.
https://github.com/moment/moment

src/pages/recentDeliveries/RecentDeliveriesTable.tsx
```diff
@@ -0,0 +1,122 @@
...
+    sortable: true,
+    filterable: true
+  }
+];
+
+export interface RecentDeliveriesTableProps {
+  data: RequestWithDate[];
+}
+
+export default class RecentDeliveriesTable extends React.Component<
```
alisman, line 103: always use react functional components.  in TS they can be defined like so:

export const MyComponent: React.FunctionComponent<MyComponentProps> = (
    props
) => {
     return <div>{..whatever}</div>
}

src/pages/requestView/RequestViewPage.tsx
```diff
@@ -1,17 +1,78 @@
...
+      var currentSm = sample.hasMetadataSampleMetadata[i];
+      if (currentSm.importDate >= latestDeliv.importDate) {
+        latestDeliv = currentSm;
+      }
+    }
+    return [firstDeliv, latestDeliv];
+  }
+}
+
+export function RequestSummary() {
```
alisman, line 44: See above this should be typed as React Function component

src/pages/recentDeliveries/RecentDeliveriesPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+    smDataList.sort();
+    return [smDataList[0], smDataList[smDataList.length - 1]];
+  }
+}
+
+function RecentDeliveriesPage() {
+  const { loading, error, data } = useQuery(ExampleQueryDocument);
+
+  function transformAndFilterRequestsByDate(requestsList: Array<Request>) {
```
alisman, line 36: helper function should be defined outside of component and then memoized using useMemo hook

src/pages/recentDeliveries/RecentDeliveriesPage.tsx
```diff
@@ -0,0 +1,71 @@
+import "./RecentDeliveries.css";
+import Container from "react-bootstrap/Container";
+import RecentDeliveriesTable from "./RecentDeliveriesTable";
+import { Request, ExampleQueryDocument } from "../../generated/graphql";
+import { useQuery } from "@apollo/client";
+
+export class RequestWithDate {
```
alisman, line 7: Using classes in js is frowned upon. You should try to use functional approach.  We can discuss this.

src/pages/recentDeliveries/RecentDeliveriesPage.tsx
ao508: 🙏

src/pages/recentDeliveries/RecentDeliveriesPage.tsx
ao508: Adding this comment as a reminder to ask about this tomorrow

src/pages/recentDeliveries/RecentDeliveriesPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+  constructor(smileRequest: Request) {
+    this.smileRequest = smileRequest;
+    this.totalSamples = smileRequest.hasSampleSamples.length;
+    let deliveryDates = this.resolveAndSortDeliveryDates(smileRequest);
+    this.firstDelivered = deliveryDates[0];
+    this.lastDateUpdated = deliveryDates[1];
+  }
+  private resolveAndSortDeliveryDates(request: Request) {
+    var smDataList: string[] = [];
+    for (var i = 0; i < request.hasSampleSamples.length; i++) {
```
alisman, line 21: basically we want to never use native looping.  use map, filter, reduce or lodash funcions

src/pages/home/HomePage.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from "react";
+import Container from "react-bootstrap/Container";
+
+function HomePage() {
```
alisman, line 4: functional component.   type this.

PR #22, opened 2022-09-20
Newmaster

alisman:


src/lib/CSVExport.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function CSVGenerate(requests) {
+  const csvString = [
+    StaticTableColumns.map(item => item.label),
+    requests.map(req => StaticTableColumns.map(col => req[col.dataKey!]))
+  ]
+
+    .map(e => e.join("\t"))
+    .join("\n");
+
```
ao508, line 12: pending some updates from shalabh to get the export working as desired

additionally, would you mind adding a prompt for the user for where to save the file and name it?

ao508, line 12: pending some updates from shalabh to get the export working as desired

additionally, would you mind adding a prompt for the user for where to save the file and name it?

src/pages/requestView/RequestSummary.tsx
```diff
@@ -1,146 +1,153 @@
...
+  //         offset: startIndex,
+  //         limit: stopIndex
+  //       }
+  //     }
+  //   });
+  // }
+
+  // function isRowLoaded({ index }) {
+  //   return index < data.requests[-].length;
+  // }
```
ao508, line 45: is this commented out for later?

if we're moving away from needing this then we can just remove it

src/shared/components/SmileNavBar.tsx
```diff
@@ -0,0 +1,271 @@
...
+        </a>
+        {/*<i className="bi bi-list toggle-sidebar-btn"></i>*/}
+      </div>
+
+      {/*<div className="search-bar">*/}
+      {/*  <form className="search-form d-flex align-items-center" method="POST" action="#">*/}
+      {/*    <input type="text" name="query" placeholder="Search" title="Enter search keyword" />*/}
+      {/*      <button type="submit" title="Search"><i className="bi bi-search"></i></button>*/}
+      {/*  </form>*/}
+      {/*</div>*/}
```
ao508, line 21: i think we can remove this also

ao508 (changes requested):

src/lib/CSVExport.ts
alisman: We can add a custom prompt for the file name but the save location is feature of the browser that we can't configure/force.  It just depends how user has it set up (e.g. always just download to Downloads folder)

.prettierignore
```diff
@@ -30,3 +30,6 @@
 LICENSE
 packages/**/dist/**
 yarn.lock
+
```
ao508, line 33: based on the output from `prettier` I think we'll need to add `*.graphql` to this file

ao508 (changes requested):
Need to check in the modified files that `prettier` generates for the following (based on the circleci test output):

- src/generated/graphql.tsx
- src/index.css
- src/pages/requestView/RequestSummary.tsx

PR #55, opened 2022-12-20
Dev sample revisable

ao508:


server/index.js
```diff
@@ -169,27 +141,22 @@ async function main() {
...
-    // cant seem to be able to query successfuly for some reason even though the
-    // status in the database is being updated to whatever I'm sending to the api
-    // see docs here https://neo4j.com/docs/graphql-manual/current/ogm/examples/custom-resolvers/
-    console.log("querying by: ", req.body.requestId);
-    const { r } = await Request.find({
-      where: {
-        igoRequestId: req.body.requestId
+  // custom mutation resolver for updating the sample revisable status to false
+  app.get("/mutation/sample/:smileSampleId", async (req, res) => {
+    const response = await server.executeOperation({
```
ao508, line 146: on the update sample mutation query:
1. validate that it's revisable (not locked)
2. If revisable, flip it into revisable=false to lock it
3. Send out Nats message

ao508: for the NATS message: unpack from request body

server/resolvers.js
```diff
@@ -0,0 +1,23 @@
...
+
+exports.resolvers = {
+  Mutation: {
+    sampleRevisableMutation: async (_, { smileSampleId, revisable }) => {
+      return updateSamples;
+    }
+  }
+};
+
+exports.sampleRevisableMutationQuery = gql`
```
alisman, line 11: this should actually accept all of the sample properties which the user is updating.  perhaps even just all the sample properties.   it doesn't know about revisable.

server/index.js
alisman: this get hander can go away completely because it's being handled by graphql

PR #56, opened 2022-12-20
Init commit resolver in dashboard frontend

ao508:


src/pages/requests/helpers.tsx
```diff
@@ -36,6 +39,75 @@ export function buildRequestTableColumns(navigate: any): ColDef[] {
...
+    }
+  });
+
+  return (
+    <Button
+      variant="outline-secondary"
+      size="sm"
+      onClick={() => {
+        // refresh data
+        let r = sampleRevisableMutationMutation()
```
alisman, line 64: coupla things here.  i wouldn't worry about checking the revisable here in the frontend on click of the button.  that needs to be checked in the resolver on the server.  if revisable is false there, then the resolver should throw/return an error.  

also, you may just not have gotten to this yet but we should be submitting the edited sample properties here, not revisable property.  changing revisable is the responsibility of the resolver.

PR #60, opened 2022-12-22
Migrate graphql server to typescript and integrate sample update changes

alisman:


ao508 (changes requested):
notes:
- had to install tsc and ts-node-dev before building to resolve error when building the backend

Changes required for `graphql-server/Dockerfile`:

I think they should be updated to:

```
 FROM node:16.10.0
 ADD ./ /server
 WORKDIR /server
RUN yarn build:backend
CMD ["yarn", "run", "start:backend"]
```

and the `frontend/Dockerfile` should be updated to:

```
FROM node:16.10.0
ADD ./ /smile-dashboard
WORKDIR /smile-dashboard
RUN yarn install
CMD ["yarn", "run", "dev:frontend"]
```

does that seem right?

PR #75, opened 2023-04-05
Query to retrieve a list of metadata by input Id

divyamadala30:


ao508 (changes requested):
Can you add the query to the `operations.graphql` file?

Once that's in there you should change your `codegen.yml` file graphql URL setting to the dev server (from localhost) and then run 

```
yarn run codegen
```

This command will update the `generated/graphql.ts` files for the dashboard frontend and backend

PR #83, opened 2023-05-03
Enable column autosizing

qu8n:
This PR enables column autosizing following [this issue](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/936).

Notes:

This implementation followed the [AG Grid doc](https://www.ag-grid.com/javascript-data-grid/column-sizing/)'s recommendation of setting `columnApi.autoSizeColumns()` after the grid is ready, via onFirstDataRendered. However, `autoSizeColumns()` only gets applied to columns/rows that are visible to users. To get around this, the AG Grid doc recommends turning off its column virtualization effect, which is something that we had discussed not wanting to do since it's hacky and could affect performance.

My solution used a workaround found [here](https://github.com/ag-grid/ag-grid/issues/1692), which involves (1) squeezing all columns to fit them into the screen via `sizeColumnsToFit()` when the grid first loads, then (2) running `autoSizeAllColumns()`. This way, all columns get autosized and not just the visible ones. Note that with this approach, the rows below the viewport are not considered, so there might be cells with enough data to the point where their contents get truncated. To cover these scenarios, I enabled all columns to be resizable.

### DIFF
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -15,8 +15,8 @@ import "ag-grid-enterprise";
 import { ColDef, IServerSideGetRowsParams } from "ag-grid-community";
 import { useHookGeneric } from "../shared/types";
 import { SamplesList } from "./SamplesList";
-import { SampleMetadataWhere, SampleWhere } from "../generated/graphql";
-import { defaultColDef } from "../pages/requests/helpers";
+import { SampleWhere } from "../generated/graphql";
+import { defaultRecordsColDef } from "../pages/requests/helpers";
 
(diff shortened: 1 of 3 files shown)


qu8n:
Looks like this commit was not signed. I reset my laptop recently and haven't configured a signature for my local commits yet. Let me configure it and override this commit.

ao508 (approved):

0xrasla:
onGridSizeChanged={(e) => {
          e.api.sizeColumnsToFit();
        }}
        
i fixed it by using this! thanks for the idea

PR #86, opened 2023-05-16
Combine Revisable and Status columns into one

qu8n:
This PR contains the work for [this issue](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/951). We're combining the Revisable and Status columns into one to make information more relevant and concise for the PMs.

Desired behavior for the combined column:

- if revisable = false, show "loading" icon
- if revisable = true and validationStatus = true, show checkmark
- if revisable = true and validationStatus = false, show warning icon

### DIFF
--- a/frontend/src/pages/requests/helpers.tsx
+++ b/frontend/src/pages/requests/helpers.tsx
@@ -194,58 +194,35 @@ export const SampleDetailsColumns: ColDef<SampleMetadataExtended>[] = [
     editable: (params) => !protectedFields.includes(params.colDef.field!),
   },
   {
-    field: "revisable",
-    headerName: "Revisable",
-    cellRenderer: function (
-      params: EditableCallbackParams<SampleMetadataExtended>
-    ) {
-      return params.data?.revisable ? (
-        <span>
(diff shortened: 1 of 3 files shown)


ao508 (changes requested):
On editing a sample / submitting changes, the record gets highlighted red (expected) but the loading ring icon does not display

frontend/src/pages/requests/helpers.tsx
```diff
@@ -472,4 +448,5 @@ const protectedFields: string[] = [
   "species",
   "validationStatus",
   "validationReport",
+  // "status",
```
ao508, line 451: let's add status back to the protected fields list (if the row is a checkmark I'm allowed to double click and see its value is actually `true`)

ao508: actually i see that it's actually the "revisable" field that is editable. let's add revisable to the list here instead and remove status.

ao508 (changes requested):
minor changes requested. otherwise looks good!

PR #94, opened 2023-06-27
Add Samples page

qu8n:
This PR contains the work for [issue #990](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/990), which involves creating a new Samples page for displaying and searching SampleMetadata.
- This Samples page currently does not implement infinite scrolling like Requests and Samples pages as that would introduce additional refactoring. We may perform this refactoring in the future as needed.
- To avoid querying the entire SampleMetadata set, this Samples page always limits the query results to 500 rows. Note that users can search through the entire SampleMetadata set; only the query results are limited.
- When users scroll to the last row, they will see a modal that suggest them to refine their search.
- There is currently no sorting logic applied to the queried SampleMetadata. This is because meaningful fields for sorting (e.g. `importDate`) are all nested fields, and [GraphQL doesn't seem to natively support sorting by nesting fields](https://stackoverflow.com/a/49642804). We can research workarounds if this matters to the users.


ao508 (changes requested):
image


The search param doesn't seem to be filtering results properly. 

Ex: Searching for "Normal" and hitting enter refreshes the page but the table doesn't appear to show results that reflect the search term. 

Follow up question. Did we decide that we wouldn't be supporting fuzzy searching anymore? Just wanted to note that this doesn't appear to be working anymore but couldn't remember what the decision was.

Note that this review isn't complete yet but I just wanted to point these things out before adding more comments.

qu8n:
> image
> 
> The search param doesn't seem to be filtering results properly.
> 
> Ex: Searching for "Normal" and hitting enter refreshes the page but the table doesn't appear to show results that reflect the search term.
> 
> There also seems to be a lot of latency between when you enter a single search term and the page response even when hitting the `Enter` key
> 
> Follow up question. Did we decide that we wouldn't be supporting fuzzy searching anymore? Just wanted to note that this doesn't appear to be working anymore but couldn't remember what the decision was.
> 
> Note that this review isn't complete yet but I just wanted to point these things out before adding more comments.

@ao508 Thanks. Let me look into the filtering issue. Looks like this has been an existing issue for the general Samples component as I'm seeing it on the Requests/Samples pages as well.

Re: the latency, do you experience a similar loading speed on the Patients page, where there are 100k+ rows? My guess is this is simply due to the sheer number of rows to search through.

Re: fuzzy search, this is what we last settled on: (1) exact match when there are multiple values and (2) fuzzy match when there is a single value.

ao508:
> > image
> > The search param doesn't seem to be filtering results properly.
> > Ex: Searching for "Normal" and hitting enter refreshes the page but the table doesn't appear to show results that reflect the search term.
> > There also seems to be a lot of latency between when you enter a single search term and the page response even when hitting the `Enter` key
> > Follow up question. Did we decide that we wouldn't be supporting fuzzy searching anymore? Just wanted to note that this doesn't appear to be working anymore but couldn't remember what the decision was.
> > Note that this review isn't complete yet but I just wanted to point these things out before adding more comments.
> 
> @ao508 Thanks. Let me look into the filtering issue. Looks like this has been an existing issue for the general Samples component as I'm seeing it on the Requests/Samples pages as well.
> 
> Re: the latency, do you experience a similar loading speed on the Patients page, where there are 100k+ rows? My guess is this is simply due to the sheer number of rows to search through.
> 
> Re: fuzzy search, this is what we last settled on: (1) exact match when there are multiple values and (2) fuzzy match when there is a single value.

I was only using one term for searching so it seems that the fuzzy match is broken at the moment

PR #95, opened 2023-07-19
Implement bulk searching for Requests page

qu8n:
See the related Zenhub issue [here](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/976). This PR implements the following changes:

- Enabling bulk searching for the Requests page: 
    - Fuzzy match if the search input contains a single value, and exact match for multiple values
    - Multiple values can be separated by either spaces or commas
    - Other input sanitization logics include removal of leading/trailing white spaces, empty values, and duplicates 
- Changing searching behavior:
    - Before: search gets triggered live as the user types in the search bar
    - After: search only starts after the user presses "Enter" or click the "Search" button
 - Making the word "requests" in "___ matching requests" singular when there is a single result
 - Increasing request limit of Express server's `bodyParser` middleware to allow for bulk searching of 100+ values

I tested this locally using new Neo4j data exported from prod with Ben's help.

### DIFF
--- a/frontend/src/lib/parseSearchQueries.ts
+++ b/frontend/src/lib/parseSearchQueries.ts
@@ -1,15 +1,7 @@
 export function parseSearchQueries(value: string): string[] {
   const queries = value
     .split(/[\s,]+/) // split on whitespaces and commas
-    .map((query) => {
-      query = query.trim();
-      // Handle cases where users bulk search by copying and pasting from a spreadsheet,
-      // and leading zeros are stripped from a IGO Request ID (e.g. `01234` becomes `1234`)
-      if (query && query.length < 5) {
-        query = query.padStart(5, "0");
-      }
-      return query;
-    })
+    .map((query) => query.trim())
     .filter(Boolean); // remove empty strings (e.g. when `value` is `123,,456`)
   const uniqueQueries = Array.from(new Set(queries));
   return uniqueQueries;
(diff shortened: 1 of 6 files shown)


qu8n:
@ao508 Tooltip preview:
CleanShot 135851@2x

frontend/src/lib/parseSearchQueries.ts
```diff
@@ -0,0 +1,8 @@
+export function parseSearchQueries(value: string): string[] {
+  const queries = value
+    .split(/[\s,]+/) // split on whitespaces and commas
+    .map((query) => query.trim())
+    .filter(Boolean); // remove empty strings (e.g. when `value` is `123,,456`)
```
alisman, line 5: i would avoid doing "tricks" like this.  instead, use lodash library which provides nicely named methods for these operations.   e.g. compact (removes falsy values) and uniq which will make the list unique.

qu8n: Thanks, @alisman. I'll look into these methods. Considering that this PR has been merged, I'll incorporate your feedback in a separate PR.

alisman: yeah, this is not too important since what you have works.  but it's good to get in the habit of avoiding these kind of things because it's hard to read.

qu8n: That makes sense. Conveniently, I'm currently working on a PR that modifies this function, so I'll incorporate your feedback there and add you as another reviewer. Thanks again.

PR #97, opened 2023-07-28
Implement bulk searching for Patients view

qu8n:
[Related Zenhub issue.](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/977)

This PR:
- implements bulk searching for the Patients page in a similar way as the Requests page
- also enables multi-cell selection for selecting multiple IDs (only for `RecordsList` because `SamplesList` already has this enabled)

I tested this locally with the dev server database. Performance should not be an issue: I bulk searched 5,000 values and got the results back in ~3s, which is similar to the initial page load.


ao508 (commented):
One slightly related request - can you add back the smilePatientId column but just have it hidden by default?

PR #98, opened 2023-08-02
Implement bulk searching for Samples view

qu8n:
[Related ZenHub issue](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/978).

This PR introduces bulk searching for both the standalone Samples page and the Samples modal within the Requests and Patients pages. 

Note that table data on the Requests and Patients pages are rendered server-side, while those on the Samples view are rendered client-side. AG Grid has different loading animations for each rendering style (see images below). It doesn't look like we can use the same loading animation for all these pages, unless they use the same rendering style.

Samples page's loading:
CleanShot 133207

Requests page's loading:
CleanShot 133120

### DIFF
--- a/frontend/src/components/SamplesList.tsx
+++ b/frontend/src/components/SamplesList.tsx
@@ -7,7 +7,7 @@ import {
 import AutoSizer from "react-virtualized-auto-sizer";
 import { Button, Col, Form, Row } from "react-bootstrap";
 import classNames from "classnames";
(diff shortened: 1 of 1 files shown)


qu8n:
@ao508 These changes have been pushed to the dev server.

ao508 (approved):

PR #100, opened 2023-08-17
Search with MRNs

qu8n:
This PR contains my latest in-progress work on MRN searching. Feel free to add your changes to this PR or another PR altogether for your rework.

### Current workflow from the user's perspective:
1. Enter MRN(s) into the search bar on the Patients page
2. Toggle on the `Search with MRNs` switch
3. Press Enter or click `Search`
4. Get back filtered results along with a new MRN column that shows MRNs corresponding to each CMO Patient ID and (if applicable) DMP Patient ID

This solves the PMs' need for a MRN to CMO/DMP Patient ID mapping.

### How it works under the hood:
1. After step 3 above, the MRN value(s) get passed into `parseSearchQueries()` to be parsed/cleaned up
2. The frontend calls an external API and passes these MRNs to it. This API then returns an array of MRN-CMO-DMP triplet(s). For example, the API would return the JSON below if we pass two MRNs to it (actual IDs redacted)
```
[
    {
        "dmpId": "",
        "cmoId": "",
        "ptMrn": ""
    },
    {
        "dmpId": "",
        "cmoId": "",
        "ptMrn": ""
    }
]
```
3. The frontend creates an array of only `cmoId`s found from step 2 and use them to query the database like usual
4. The frontend gets back the data from the database, cross-checks the array of triplet(s) found from step 2, and adds a new MRN column to the AG Grid table that shows the MRN corresponding to each pair of CMO and DMP IDs

### Misc. notes:
- The external API mentioned above can be found [here](https://github.com/mskcc/smile-server/pull/1015). Please reach out to Ben for help setting this API server up locally
- The `Search with MRNs` toggle was added because there isn't a specific pattern of MRNs we can write logic to recognize

Related Zenhub issue: [link](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/board)

### DIFF
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -34,6 +34,7 @@ export interface IRecordsListProps {
   searchVariables: SampleWhere;
   customFilterUI?: JSX.Element;
   customFilterState?: boolean;
+  setCustomFilterVals?: (vals: any[]) => void;
 }
 
 const RecordsList: FunctionComponent<IRecordsListProps> = ({
@@ -49,6 +50,7 @@ const RecordsList: FunctionComponent<IRecordsListProps> = ({
   searchVariables,
   customFilterUI,
   customFilterState,
+  setCustomFilterVals,
 }) => {
   const [val, setVal] = useState("");
   const [searchVal, setSearchVal] = useState<string[]>([]);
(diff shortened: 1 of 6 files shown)


qu8n:
Please reference PR #101 for more up-to-date work on this effort.

PR #101, opened 2023-10-05
MRN search and auth implementation

qu8n:
This PR implements an MRN search toggle UI and authentication when in "MRN mode". 

Flow:

- [x] User visits SMILE dashboard
- [x] User enables MRN search mode (via the toggle)
- [x] User enters MRN inputs in search bar and clicks Search
- [x] Frontend makes an HTTP POST request to the backend’s /mrn-search API endpoint + includes the MRN inputs in its request body
- [x] Backend receives the request and checks if the user is authenticated
    - If yes: returns data with status 200
    - If no: returns status 401
- [x] Frontend sees 401 and redirects user to a popup window to login with Keycloak
- [x] User authenticates successfully
- [x] Popup window messages the original window to continue with MRN search then closes, user is back to the original window
- [x] Frontend still retains the MRN inputs in the search bar and automatically repeats step 4. The /mrn-search API endpoint now returns status 200


frontend/src/App.tsx
```diff
@@ -0,0 +1,35 @@
...
+            <Route path=":cmoPatientId" />
+          </Route>
+          <Route path="/samples" element={<SamplesPage />} />
+        </>
+      </Routes>
+    </main>
+  );
+}
+
+export default App;
```
qu8n, line 35: Most of this code and routes were directly moved out of `frontend/src/index.tsx`. I moved them into App, a functional component, to use hooks. 

Purpose of the `userEmail` state: After user logs in for the first time, the login popup window would post a message to the opener window that includes the user's email. That way, the opener window's navbar can immediately display "Logged in as {userEmail}" and a logout button. Work to persist this will come in a later ticket.

frontend/src/components/RecordsList.tsx
```diff
@@ -27,10 +28,14 @@ export interface IRecordsListProps {
...
   colDefs: ColDef[];
-  conditionBuilder: (val: string) => Record<string, any>[];
+  conditionBuilder: (uniqueQueries: string[]) => Record<string, any>[];
   sampleQueryParamValue: string | undefined;
   sampleQueryParamFieldName: string;
   searchVariables: SampleWhere;
+  customFilterUI?: JSX.Element;
+  customFilterState?: boolean;
+  setCustomFilterVals?: (vals: any[]) => void;
+  customFilterFunc?: (vals: any[]) => Promise<string[]>;
```
qu8n, line 38: These new props allow us to agnostically pass in the MRN search toggle widget and related functionalities. See `frontend/src/pages/patients/PatientsPage.tsx` to see how these props are being used.

frontend/src/components/RecordsList.tsx
```diff
@@ -120,6 +129,21 @@ const RecordsList: FunctionComponent<IRecordsListProps> = ({
...
+      setCustomFilterVals!.length > 0 &&
+      customFilterFunc
+    ) {
+      const newQueries = await customFilterFunc(uniqueQueries);
+      setSearchVal(newQueries);
+    } else {
+      setSearchVal(uniqueQueries);
+    }
+  };
+
```
qu8n, line 146: How this works in a nutshell:
- If the user is performing a standard search/filter (with MRN mode off), then everything carries on as before
- However, if they're searching with MRNs, then we'll fetch from the protected /mrn-search API endpoint, receive and parse out the corresponding CMO Patient IDs, then use those CMO Patient IDs to search/filter as normal

frontend/src/lib/CSVExport.ts
```diff
@@ -2,7 +2,7 @@ import { ColDef } from "ag-grid-community";
 
 export function CSVFormulate(rows: any[], columnDefinitions: ColDef[]) {
   const csvString = [
-    columnDefinitions.map((item) => item.field).join("\t"),
+    columnDefinitions.map((item) => item.headerName).join("\t"),
```
qu8n, line 5: This is a cosmetic fix unrelated to MRN search so that exported spreadsheet headers will use the custom `headerName` instead of `field`'s value.

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,9 +71,101 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+    }
+
+    if (popupOpen) {
+      window.addEventListener("message", handleLogin);
+      return () => {
+        window.removeEventListener("message", handleLogin);
+      };
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [popupOpen]);
```
qu8n, line 99: See my comment [here](https://github.com/mskcc/smile-dashboard/pull/101/files#r1367406307) for more info.

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,9 +71,101 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+          };
+        } else {
+          return column;
+        }
+      });
+      setPatientsListColumns(colsWithMRNs);
+    } else {
+      setPatientsListColumns(PatientsListColumns);
+    }
+  }, [searchWithMRNs, patientIdsTriplets]);
```
qu8n, line 128: In MRN search mode, this `useEffect` helps show and populate a new column with MRN values that correspond to CMO and DMP Patient IDs.

graphql-server/src/index.ts
```diff
@@ -34,19 +37,167 @@ const driver = neo4j.driver(
...
 
+// OracleDB requires node-oracledb's Thick mode & the Oracle Instant Client, which is unavailable for M1 Macs
+let oracledb: any = null;
+const os = require("os");
+if (os.arch() !== "arm64") {
+  oracledb = require("oracledb");
+  oracledb.initOracleClient();
+  oracledb.outFormat = oracledb.OUT_FORMAT_OBJECT;
+}
+
```
qu8n, line 48: Please ignore OracleDB-related code for this round of the review.

frontend/src/components/RecordsList.tsx
```diff
@@ -27,289 +28,331 @@ export interface IRecordsListProps {
...
   searchTerm: string;
   colDefs: ColDef[];
-  conditionBuilder: (val: string) => Record<string, any>[];
+  conditionBuilder: (uniqueQueries: string[]) => Record<string, any>[];
   sampleQueryParamValue: string | undefined;
   sampleQueryParamFieldName: string;
   searchVariables: SampleWhere;
+  customFilterUI?: JSX.Element;
+  customFilterState?: boolean;
+  setCustomFilterVals?: (vals: any[]) => void;
```
alisman, line 37: we should avoid use of any.

frontend/src/components/RecordsList.tsx
```diff
@@ -27,289 +28,331 @@ export interface IRecordsListProps {
...
-  sampleQueryParamFieldName,
-  searchVariables,
-}) => {
-  const [val, setVal] = useState("");
-  const [searchVal, setSearchVal] = useState("");
-  const [showDownloadModal, setShowDownloadModal] = useState(false);
-  const [showClosingWarning, setShowClosingWarning] = useState(false);
-  const [unsavedChanges, setUnsavedChanges] = useState(false);
-  const navigate = useNavigate();
+interface Ref {
```
alisman, line 41: seems an odd name for this interface.

frontend/src/index.tsx
```diff
@@ -29,34 +26,15 @@ const cache = new InMemoryCache({
 const client = new ApolloClient({
   uri:
     process.env.REACT_APP_GRAPHQL_CLIENT_URI === undefined
```
alisman, line 28: is there really env varaibles available in frontend?

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,8 +74,111 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+  const [phiEnabled, setPhiEnabled] = useState(false);
+  const [patientIdsTriplets, setPatientIdsTriplets] = useState<
+    PatientIdsTriplet[]
+  >([]);
+  const [patientsListColumns, setPatientsListColumns] =
+    useState(PatientsListColumns);
+  const [popupOpen, setPopupOpen] = useState(false);
+  const [showAlertModal, setShowAlertModal] = useState(false);
+
+  useEffect(() => {
```
alisman, line 94: this seems like duplicate code. shoud only exist in once place in app.

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,8 +74,111 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+    function handleLogin(event: any) {
+      if (event.origin !== "http://localhost:4001") return;
```
alisman, line 96: all urls should be passed to app in a global configration object.  no hard coding

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,8 +74,111 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+      setUserEmail(event.data);
+      setPopupOpen(false);
+      childRef.current.handleSearch();
+    }
+
+    if (popupOpen) {
+      window.addEventListener("message", handleLogin);
```
alisman, line 103: you can probably set up listener globally once and not worry about removing it like this.

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,8 +74,111 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+    patientMrns: string[]
+  ): Promise<string[]> {
+    try {
+      const response = await fetch("http://localhost:4001/mrn-search", {
+        method: "POST",
+        credentials: "include",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(patientMrns),
```
alisman, line 150: not urgent, but we should switch to a library that wraps fetch.  Axios library seems to be very commonly used.

alisman: this is another reason to just graphql, because it will make everything consistent, instead of having two ways of fetching and handling data.

alisman: the drawback is we have to learn a little more about graphQL. But that's not so bad!

frontend/src/index.tsx
qu8n: Yes, for local development purposes. When making changes to the frontend only, we can set this env var to the backend running on the dev server and avoid spinning up a local backend server

frontend/src/pages/patients/PatientsPage.tsx
qu8n: Agreed (see PR description). There's a bit more legwork to make all of this production-ready, which includes removing this hard coding.

> Getting this ready for production will be a separate ticket, so please ignore instances of “localhost:…“” in the code

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,8 +74,111 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+    if (popupOpen) {
+      window.addEventListener("message", handleLogin);
+      return () => {
+        window.removeEventListener("message", handleLogin);
+      };
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [popupOpen]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
qu8n: Great catch, and totally makes sense. Thank you.

frontend/src/pages/patients/PatientsPage.tsx
alisman: completely trivial!

qu8n: @alisman After more testing and research, I've decided to mostly keep this code as-is. Native DOM methods like `addEventListener` don't stay up to date with React's state variable changes ([a bit more info here](https://stackoverflow.com/questions/67244161/react-js-state-not-updating-in-event-listener-in-useeffect-hook)). Because the `handleLogin` function relies on certain React state variables, if we add this listener at the beginning of the page load, `handleLogin` won't work correctly. Happy to discuss this with you live.

frontend/src/index.tsx
qu8n: Yes. TL;DR we have a workflow to make this available in the frontend.

My understanding is that when we build the Docker images for `smile-dashboard`, we first run a bash script that exports the variable `REACT_APP_GRAPHQL_CLIENT_URI` to the host machine's environment, then our `docker-compose` file specifies adding this variable to the Docker container's own environment. 

By [default](https://create-react-app.dev/docs/adding-custom-environment-variables/), the Create React App build system can embed environment variable names with the `REACT_APP_` prefix to the application.

In the future, we'll move away from running this script and just use a `.env` file for storing environment variables.

alisman: thanks for this explaination

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,9 +76,144 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
   }
 }
 
-export const PatientsPage: React.FunctionComponent = (props) => {
+function addCDashToCMOId(cmoId: string): string {
+  return cmoId.length === 6 ? `C-${cmoId}` : cmoId;
+}
+
+const phiWarning = {
```
alisman, line 83: no biggie but it's a good practice to make "content" constants like this all caps.  e.g. PHI_WARNING.  This distinguishes them from variables that are more logic related

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -65,9 +76,144 @@ function patientAliasFilterWhereVariables(value: string): PatientAliasWhere[] {
...
+      handleSearch();
+    }
+
+    return () => {
+      window.removeEventListener("message", handleLogin);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [phiEnabled]);
+
+  PatientsListColumns = useMemo(() => {
```
alisman, line 189: this seems very odd.  you are using PatientsListColumns inside of the useMemo which is mean to create it?  
Also, a varaible like PatientsListColumns should always have a var, let or const before it.   and it should be camelcase.  capitalized is should be reserved for Reach component.

alisman: just to be clear, useMemo should definately not refer to itself.  it should be a pure function that computes something from its dependencies (or some known static value)

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -100,9 +246,146 @@ export const PatientsPage: React.FunctionComponent = (props) => {
...
+    field: "patientMrn",
+    headerName: "Patient MRN",
+    hide: true,
+    cellStyle: { color: "crimson" },
+  },
+  {
+    field: "cmoPatientId",
+    headerName: "CMO Patient ID",
+    valueGetter: function ({ data }) {
+      for (let i of data["isAliasPatients"][0]["patientAliasesIsAlias"]) {
```
alisman, line 343: never use native for loops.   if your iterating over an array, use array.forEach.   if object, use lodash _.forIn

alisman: actually, this loop would be more semantically expressed with array.find.  it would also let you handle case where nothing is found.

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -100,9 +246,146 @@ export const PatientsPage: React.FunctionComponent = (props) => {
...
+        }
+      }
+    },
+    sortable: false,
+  },
+  {
+    field: "dmpPatientId",
+    headerName: "DMP Patient ID",
+    valueGetter: function ({ data }) {
+      for (let i of data["isAliasPatients"][0]?.patientAliasesIsAlias) {
```
alisman, line 355: ditto

frontend/src/shared/components/SmileNavBar.tsx
```diff
@@ -19,9 +24,26 @@ const SmileNavBar: FunctionComponent = () => {
...
           <NavLink href="/samples">Samples</NavLink>
         </Nav>
+        {userEmail && (
+          <div className="ms-auto d-flex">
+            <p className="m-auto">Logged in as {userEmail}</p>
+            <button
+              type="button"
+              className="btn btn-outline-primary btn-sm m-3"
+              onClick={() => {
+                fetch("https://localhost:4000/logout", {
```
alisman, line 34: avoid embedding network request inside JSX callback.  You should instead declare a function and call it.

graphql-server/src/index.ts
```diff
@@ -39,14 +54,202 @@ async function main() {
...
+      { client: keycloakClient },
+      (tokenSet: any, userinfo: any, done: any) => {
+        ssoSessionIdleTimeout = tokenSet.refresh_expires_in * 1000; // convert to ms because it's in s by default
+        id_token = tokenSet.id_token;
+        return done(null, tokenSet.claims());
+      }
+    )
+  );
+
+  app.get("/login", (req, res, next) => {
```
alisman, line 114: this stuff will all disappear when we follow up with graphql implementation, but in general we should avoid dumping alll this stuff into the "main" file (index.ts).  Right now it's just a mishmash of things.

graphql-server/src/index.ts
```diff
@@ -39,14 +54,202 @@ async function main() {
...
+      const patientIdsTriplets = [];
+
+      if (os.arch() !== "arm64" && oracledb !== null) {
+        const connection = await oracledb.getConnection({
+          user: props.oracle_user,
+          password: props.oracle_password,
+          connectString: props.oracle_connect_string,
+        });
+
+        for (const patientMrn of patientMrns) {
```
alisman, line 229: this code can be improved using Promise.all.  We can discuss.

frontend/src/pages/patients/PatientsPage.tsx
qu8n: Let's discuss this one live. I have a few clarifying questions. Thanks, Aaron.

graphql-server/src/index.ts
qu8n: Agreed. A lot of this stuff will change with the GraphQL implementation, so I didn't want to prematurely refactor/organize this code into helper files.

graphql-server/src/index.ts
qu8n: I got a head start and thought the refactored code could look like the below. Will save this for when we chat live.

```
const patientIdsTriplets = [];

if (os.arch() !== "arm64" && oracledb !== null) {
  const connection = await oracledb.getConnection({
    user: props.oracle_user,
    password: props.oracle_password,
    connectString: props.oracle_connect_string,
  });

  const promises = patientMrns.map(async (patientMrn) => {
    const result = await connection.execute(
      "SELECT CMO_ID, DMP_ID, PT_MRN FROM CRDB_CMO_LOJ_DMP_MAP WHERE :patientMrn IN (DMP_ID, PT_MRN, CMO_ID)",
      { patientMrn }
    );
    if (result.rows.length > 0) {
      return result.rows[0];
    }
  });

  patientIdsTriplets.push(...(await Promise.all(promises)));

  await connection.close();
}
```

PR #106, opened 2024-01-19
Use GraphQL query for fetching MRN data and reorganize graphql-server code

qu8n:
This PR replaces the built-in `fetch` method that we're using to request MRN data from the frontend (`smile-dashboard`) with a GraphQL query. This change allows us to standardize our data fetching to all be done via GraphQL.

As part of this PR, I also reorganized/cleaned up the `graphql-server` codebase. We previously put almost all code into the root `index.ts` file. I'm moving middlewares, API endpoints, and servers configs into their respective folders. New comments were added and some old comments were revised to improve clarity.

@ao508 ~~This change has not been tested in dev. I'm investigating some other Podman issues. Will loop you in during our next standup.~~ 
1/25 update: this PR has been pushed to dev.

1/26 update: this PR has been rebased to the latest master commit (https://github.com/mskcc/smile-dashboard/commit/c29343ff804c863a7110705af313a5935e247764).


frontend/package.json
```diff
@@ -75,7 +75,6 @@
     ]
   },
   "devDependencies": {
-    "@graphql-codegen/cli": "2.8.1",
```
qu8n, line 78: Cleanup: when we run the `yarn run codegen` script, it looks for the `@graphql-codegen/cli` package in the root directory, but this package is currently installed in the `frontend` directory. I'm moving this package to the root directory.

frontend/src/generated/graphql.ts
qu8n, line 1: These are types auto-generated by the GraphQL Code Generator to support the new GraphQL query created to fetch MRN data. See `graphql/operations.graphql` for the query definition.

frontend/src/generated/graphql.ts
```diff
@@ -795,6 +795,13 @@ export type PatientHasSampleSamplesUpdateFieldInput = {
   where?: InputMaybe<PatientHasSampleSamplesConnectionWhere>;
 };
 
+export type PatientIdsTriplet = {
+  __typename?: "PatientIdsTriplet";
```
alisman, line 799: what does this typename correspond to?

alisman: oh this is generated, nvm

graphql-server/src/index.ts
```diff
@@ -1,315 +1,25 @@
...
-  app.get("/auth/callback", (req, res, next) => {
-    // This second passport.authenticate() serves a distinct function from the above.
-    // Lets Keycloak respond to the above authentication request, following the OpenID protocol.
-    // If successful, this adds `isAuthenticated()` to the `req` object which is used below.
-    passport.authenticate("oidc", {
-      successRedirect: "/post-login",
-      failureRedirect: "/",
-    })(req, res, next);
-  });
+  require("./middlewares/express")(app);
```
alisman, line 8: this should be imports at the top of the file, not requires.

graphql-server/src/middlewares/express.ts
```diff
@@ -0,0 +1,10 @@
+import express, { Express } from "express";
+import cors from "cors";
+import bodyParser from "body-parser";
+import { corsOptions } from "../utils/constants";
+
+module.exports = function (app: Express) {
```
alisman, line 6: you should just be able to use export keyword

graphql-server/src/middlewares/logging.ts
```diff
@@ -0,0 +1,43 @@
+import { Express } from "express";
+import path from "path";
+const morgan = require("morgan");
```
alisman, line 3: import?

graphql-server/src/middlewares/session.ts
```diff
@@ -0,0 +1,65 @@
+import { Express } from "express";
+import { Strategy } from "openid-client";
+const passport = require("passport");
+const session = require("express-session");
+import { buildProps } from "../utils/buildProps";
+import { getKeycloakClient } from "../utils/session";
+
+const props = buildProps();
```
alisman, line 8: usually best not to call a function in the module scope like this.  files should be inert.  nothing happens unless something is called

graphql-server/src/middlewares/session.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * This middleware sets up the following session-related mechanisms:
+ * - Express-session for storing user info in session store and managing the session cookie
+ * - Passport for authenticating users using OpenID Connect (OIDC) protocol with Keycloak as the OIDC provider
+ * - Storing timestamp of user's last activity for the auto-timeout functionality
+ */
+module.exports = async function (app: Express) {
```
alisman, line 16: best to give a function like this a name that describes what it does (e..g createSession)

graphql-server/src/routes/auth/logout.ts
```diff
@@ -0,0 +1,25 @@
+import { getKeycloakClient } from "../../utils/session";
+const fetch = require("node-fetch");
+
+/**
+ * Clear user from `activeUserSessions` tracker, `req` object, and their Keycloak session.
+ * User is logged out from Keycloak without a "log out?" prompt and redirecting the response.
+ */
+export async function logOutRouter(req: any) {
```
alisman, line 8: is this really a "router"?

graphql-server/src/routes/auth/post-login.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Upon successful login, the user is redirected to this route, which
+ * closes the popup window and sends user's email to the parent window
+ * to display on in the navbar.
+ */
+export function postLoginRouter(req: any, res: any) {
+  const userEmail = req.user.email;
+
+  res.send(`
```
alisman, line 11: this seems like code that should live in the frontend.  servive should just respond with some data indicating what to do.

qu8n: Thanks, Aaron. Heads up that I'm moving this work into [another ticket](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1086) to limit the scope of this PR to just the `graphql-server`.

graphql-server/src/middlewares/logging.ts
qu8n: For Angelica's context: 

Replacing `const morgan = require("morgan");` with `import morgan from "morgan";` results in the Typescript error "TS7016: Could not find a declaration file for module '<module-name>'. '<module's index.js file>' implicitly has an 'any' type. Try `npm install @types/<module-name>` if it exists or add a new declaration (.d.ts) file containing `declare module '<module-name>';`.

Yesterday, Aaron and I followed the error message's tip to "add a new declaration (.d.ts) file containing `declare module '<module-name>';". This made the error message go away.

qu8n: @alisman When I tried testing this today and run the local dev server with `yarn dev:backend`, the same error message reappeared. 

As a fix, I followed a StackOverflow post and tried including the path to the declaration file inside `tsconfig.json` but saw the same result.

I still like the idea of using `import` instead of `require` everywhere for its clean look, but I don't think there would be any added practical benefits because we're not defining the package types. 

I'd like to revisit this task if we do need to define the types in the future. For now, I'm dropping this task but happy to keep investigating if you think otherwise.

graphql-server/src/buildProps.ts
```diff
@@ -1,37 +0,0 @@
-import path from "path";
-const PropertiesReader = require("properties-reader");
-const properties = new PropertiesReader(
-  path.resolve(`${__dirname}`, "./env/application.properties")
```
ao508, line 4: for my own reference, where/how are props being resolved now?

graphql-server/src/utils/servers.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export function initializeHttpsServer(app: Express) {
+  const httpsServer = https.createServer(
+    {
+      key: fs.readFileSync(
+        path.join(__dirname, "../../../.cert/smile-dashboard-web-key.pem")
+      ),
+      cert: fs.readFileSync(
+        path.join(__dirname, "../../../.cert/smile-dashboard-web-cert.pem")
+      ),
```
ao508, line 24: I'm leaning towards moving these to a props/constants instead. 

Maybe a const for the cert filepath and props for the key and certificate filenames.

ao508 (commented):
One minor comment but this PR can be merged without resolving it.

graphql-server/src/utils/servers.ts
ao508: @alisman I'm curious what you think as well. This is a non-issue really and more of a preference. I don't think this should hold up merging the PR.

qu8n: I agree with making this not hard coded here. Will go ahead and merge this PR but won't delete it to give Aaron time to respond.

PR #117, opened 2024-04-03
Show billing-related fields on the Cohort and Cohort Sample views

qu8n:
Screenshots of the new billing fields in the Cohort and Cohort Sample views:
cohorts_view
cohort_samples_view

### DIFF
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -14,7 +14,7 @@ import { ColDef, IServerSideGetRowsParams } from "ag-grid-community";
 import { DataName, useHookLazyGeneric } from "../shared/types";
 import SamplesList from "./SamplesList";
 import { Sample, SampleWhere } from "../generated/graphql";
-import { defaultReadOnlyColDef } from "../shared/helpers";
+import { defaultColDef } from "../shared/helpers";
 import { PatientIdsTriplet } from "../pages/patients/PatientsPage";
 import { ErrorMessage, LoadingSpinner, Toolbar } from "../shared/tableElements";
 
@@ -36,7 +36,6 @@ interface IRecordsListProps {
   setShowDownloadModal: Dispatch<SetStateAction<boolean>>;
   handleDownload: () => void;
   samplesQueryParam: string | undefined;
-  samplesDefaultColDef: ColDef;
   getSamplesRowData: (samples: Sample[]) => any[];
   samplesColDefs: ColDef[];
   samplesParentWhereVariables: SampleWhere;
@@ -63,7 +62,6 @@ export default function RecordsList({
   setShowDownloadModal,
   handleDownload,
   samplesQueryParam,
-  samplesDefaultColDef,
   getSamplesRowData,
   samplesColDefs,
   samplesParentWhereVariables,
@@ -214,7 +212,6 @@ export default function RecordsList({
                 <div className={styles.popupHeight}>
                   <SamplesList
                     columnDefs={samplesColDefs}
-                    defaultColDef={samplesDefaultColDef}
                     getRowData={getSamplesRowData}
                     parentWhereVariables={samplesParentWhereVariables}
                     refetchWhereVariables={samplesRefetchWhereVariables}
@@ -260,7 +257,7 @@ export default function RecordsList({
               context={{
                 navigateFunction: navigate,
               }}
-              defaultColDef={defaultReadOnlyColDef}
+              defaultColDef={defaultColDef}
               onGridReady={(params) => {
                 params.api.sizeColumnsToFit();
               }}
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -587,14 +587,17 @@ export const CohortSampleDetailsColumns: ColDef[] = [
     headerName: "Initial Pipeline Run Date",
   },
   {
+    field: "billed",
     headerName: "Billed",
     editable: true,
   },
   {
+    field: "costCenter",
     headerName: "Cost Center",
     editable: true,
   },
   {
+    field: "billedBy",
     headerName: "Billed By",
   },
   {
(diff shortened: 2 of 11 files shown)


frontend/src/components/RecordsList.tsx
qu8n, line 1: Both `defaultReadOnlyColDef` and `defaultEditableColDef` were consolidated into `defaultColDef` (see its definition [here](https://github.com/mskcc/smile-dashboard/pull/117/files#diff-db966dd79103f32fddb49dcefcfba78d4325506ce364cd0e24779a3521aa7173R668)). These variables contain properties that our data tables share by default (like `sortable: true`).

We used to have separate "read-only" and "editable" versions because initially we planned to have the Cohort Sample view completely non-editable while the other Sample views editable. This is no longer the case with the addition of billing fields.

frontend/src/shared/helpers.tsx
```diff
@@ -235,6 +236,39 @@ export const PatientsListColumns: ColDef[] = [
...
+  },
+  cellRendererParams: {
+    colDef: {
+      tooltipComponent: StatusTooltip,
+      tooltipValueGetter: (params: ITooltipParams) =>
+        params.data.hasStatusStatuses[0]?.validationReport ??
+        params.data.hasStatusStatuses,
+    },
+  },
+};
```
qu8n, line 270: This change does two things:
1. Put these column definitions into a named variable `statusCellRendererProps` so that we can pass it into the Status column definition of both the Cohort Sample tables and regular Sample tables
2. Refactor some codes so that we can more easily read and understand when certain icon is used (i.e. check icon, warning icon, and loading icon)

frontend/src/shared/helpers.tsx
```diff
@@ -433,56 +438,58 @@ const readOnlyHeader = {
...
-      const changedValue = changes?.find((change: any) => {
-        return (
-          change.fieldName === params.colDef.field &&
-          change.primaryId === params.data.primaryId
-        );
-      });
-      return changedValue !== undefined;
-    },
-  };
+function setupEditableSampleFields(samplesColDefs: ColDef[]) {
```
qu8n, line 441: The codes below used to be applied to just the regular, non-Cohort Sample views to support the edit-ability of fields. Now that Cohort Sample view is also editable, I put these codes into a named function (`setupEditableSampleFields`) so that we can apply them to all sample views.

frontend/src/shared/helpers.tsx
```diff
@@ -517,16 +524,29 @@ export const CohortsListColumns: ColDef[] = [
...
+    headerName: "Billed",
+    valueGetter: ({ data }) => {
+      return data["hasCohortSampleSamples"]?.every((sample: Sample) => {
+        return sample.hasTempoTempos?.[0]?.billed === true;
+      });
+    },
+    valueFormatter: (params) => (params.value === true ? "Yes" : "No"),
+  },
+  {
+    headerName: "Initial Cohort Delivery Date",
```
qu8n, line 536: This new column name came from Pavitra today

frontend/src/shared/helpers.tsx
```diff
@@ -617,35 +662,30 @@ export const CohortSamplesDetailsColumns: ColDef[] = [
...
+  "preservation",
+  "tumorOrNormal",
+  "sampleClass",
+  "oncotreeCode",
+  "collectionYear",
+  "sampleOrigin",
+  "tissueLocation",
+  "sex",
+  "billed",
+  "costCenter",
```
qu8n, line 688: Instead of specifying fields that are non-editable, I switched this to a list of fields that are editable. This should (1) help us avoid accidentally making fields editable when they are to be protected and (2) result in less frontend codes.

frontend/src/shared/helpers.tsx
```diff
@@ -752,6 +798,12 @@ export function cohortSampleFilterWhereVariables(
     ];
   } else {
     tempoWhere = [
+      {
+        billedBy_CONTAINS: parsedSearchVals[0],
+      },
+      {
+        costCenter_CONTAINS: parsedSearchVals[0],
+      },
```
qu8n, line 806: Here I'm specifying new fields that users can search by. Note that `billed` isn't included here because (1) it is a boolean field, which doesn't support suffices `_IN` and `_CONTAINS` by GraphQL, and (2) I don't expect PMs to search for specific billed value. If they do, we could consider writing a custom resolver or adding a filter/sort ability to the Billed column in the future.

frontend/src/shared/helpers.tsx
```diff
@@ -889,3 +947,7 @@ export function handleSearch(
   const parsedSearchVals = parseUserSearchVal(userSearchVal);
   setParsedSearchVals(parsedSearchVals);
 }
+
+function formatCohortRelatedDate(date: string) {
+  return date?.split(" ")[0];
+}
```
qu8n, line 953: These changes would remove the hour and minute from the timestamps. This is a simple change so I figured I could squeeze them into this PR.

graphql/operations.graphql
```diff
@@ -94,6 +94,9 @@ query FindSamplesByInputValue(
           }
         }
         hasTempoTempos {
+          billed
+          billedBy
+          costCenter
```
qu8n, line 99: For displaying these fields on the Cohort Samples view

graphql/operations.graphql
```diff
@@ -246,5 +249,11 @@ query CohortsList(
     hasCohortSampleSamplesConnection {
       totalCount
     }
+    hasCohortSampleSamples {
+      smileSampleId
+      hasTempoTempos {
+        billed
+      }
+    }
```
qu8n, line 257: For calculating the `Billed` value on the Cohorts view

frontend/src/shared/helpers.tsx
ao508: Is this something we can support purely on the frontend where a user can filter the presented table for cohorts that have not been billed yet?

qu8n: Yes. AG Grid supports this ability ([example](https://ag-grid.com/_astro/open-column-dark.551e6e75.gif)). This feature doesn't seem like a one-line change, though, so if we're moving forward with this option I think we should create a separate ticket

ao508: Sounds good. Would you mind making that a ticket?

ao508 (approved):

PR #119, opened 2024-04-23
Fix: duplicate records on Patients page

qu8n:
This PR modifies the GraphQL query used on Patients page to fetch for `Patient` nodes instead of `PatientAlias`, which was what led to the duplicative rows.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -36,7 +36,7 @@ export default function App() {
               <PatientsPage userEmail={userEmail} setUserEmail={setUserEmail} />
             }
           >
-            <Route path=":cmoPatientId" />
+            <Route path=":smilePatientId" />
           </Route>
           <Route path="/samples" element={<SamplesPage />} />
           <Route
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -28,7 +28,6 @@ import { ErrorMessage, LoadingSpinner, Toolbar } from "../shared/tableElements";
 interface IRecordsListProps {
   colDefs: ColDef[];
   dataName: DataName;
-  nodeName?: string;
   lazyRecordsQuery: typeof useHookLazyGeneric;
   lazyRecordsQueryAddlVariables?: Record<string, any>;
   prepareDataForAgGrid?: (data: any) => any;
@@ -60,7 +59,6 @@ interface IRecordsListProps {
 export default function RecordsList({
   colDefs,
   dataName,
-  nodeName = dataName,
   lazyRecordsQuery,
   lazyRecordsQueryAddlVariables,
   prepareDataForAgGrid,
@@ -98,7 +96,7 @@ export default function RecordsList({
       },
     });
 
-  const totalCountNodeName = `${nodeName}Connection`;
+  const totalCountNodeName = `${dataName}Connection`;
 
   const datasource: IServerSideDatasource = useMemo(() => {
     return {
@@ -108,7 +106,7 @@ export default function RecordsList({
           where: {
             OR: queryFilterWhereVariables(parsedSearchVals),
           },
-          [`${nodeName}ConnectionWhere2`]: {
+          [`${dataName}ConnectionWhere2`]: {
             OR: queryFilterWhereVariables(parsedSearchVals),
           },
           options: {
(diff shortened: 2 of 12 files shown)


frontend/src/App.tsx
```diff
@@ -36,7 +36,7 @@ export default function App() {
               <PatientsPage userEmail={userEmail} setUserEmail={setUserEmail} />
             }
           >
-            <Route path=":cmoPatientId" />
+            <Route path=":smilePatientId" />
```
qu8n, line 39: This tells React that the `id` in route `/patients/id` is a `smilePatientId`. We can't keep `cmoPatientId` anyway because we also want to handle `dmpPatientId`.

frontend/src/components/RecordsList.tsx
```diff
@@ -10,20 +10,27 @@ import styles from "./records.module.scss";
...
+  defaultColDef,
+  prepareSampleMetadataForAgGrid,
+} from "../shared/helpers";
 import { PatientIdsTriplet } from "../pages/patients/PatientsPage";
 import { ErrorMessage, LoadingSpinner, Toolbar } from "../shared/tableElements";
 
 interface IRecordsListProps {
   colDefs: ColDef[];
   dataName: DataName;
-  nodeName?: string;
```
qu8n, line 24: This prop was here basically to handle the "Patients" page querying data by a node of a different name ("PatientAlias"), unlike the "Requests" and "Cohorts" pages. Now that the "Patients" page queries data of node "Patients", we no longer need this prop.

frontend/src/components/RecordsList.tsx
```diff
@@ -10,20 +10,27 @@ import styles from "./records.module.scss";
...
   lazyRecordsQuery: typeof useHookLazyGeneric;
   lazyRecordsQueryAddlVariables?: Record<string, any>;
+  prepareDataForAgGrid?: (data: any) => any;
```
qu8n, line 33: This new prop is a function that takes the data returned by the GraphQL query and transforms it to be fed directly into the AG Grid table. Previously, we were doing this for Patients data inside its AG Grid column definitions.

This approach has two benefits: (1) matching how we handle samples data/views and (2) more performant.

frontend/src/pages/cohorts/CohortsPage.tsx
```diff
@@ -158,7 +158,7 @@ export default function CohortsPage({
             },
           } as SampleWhere
         }
-        samplesRefetchWhereVariables={(samplesParsedSearchVals: string[]) => {
+        samplesRefetchWhereVariables={(samplesParsedSearchVals) => {
```
qu8n, line 161: Specifying the type here was redundant because it's inferred automatically by the IDE based on `IRecordsListProps` in `RecordsList.tsx`.

frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -29,51 +29,35 @@ export type PatientIdsTriplet = {
...
       {
-        isAliasPatients_SOME: {
-          hasSampleSamples_SOME: {
-            hasMetadataSampleMetadata_SOME: {
-              primaryId_CONTAINS: parsedSearchVals[0],
-            },
+        hasSampleSamples_SOME: {
+          hasMetadataSampleMetadata_SOME: {
+            cmoSampleName_CONTAINS: parsedSearchVals[0],
           },
```
qu8n, line 61: This function gets called when the user performs a search. Since we're querying by `Patients` now instead of `PatientAlias`, I'm updating where GraphQL searches for data.

frontend/src/shared/helpers.tsx
```diff
@@ -140,18 +140,69 @@ export const RequestsListColumns: ColDef[] = [
...
+      smilePatientId,
+    };
+  });
+
+  return {
+    patientsConnection,
+    patients: newPatients,
+  };
+}
+
```
qu8n, line 193: See [this comment above](https://github.com/mskcc/smile-dashboard/pull/119/files#r1576559456).

frontend/src/shared/helpers.tsx
qu8n, line 1: See [this comment](https://github.com/mskcc/smile-dashboard/pull/119/files#r1576559456) for context on the changes here.

qu8n, line 1: See [this comment](https://github.com/mskcc/smile-dashboard/pull/119/files#r1576559456) for context on most of the changes here.

ao508 (approved):

PR #122, opened 2024-04-30
Add embargo date to cohort samples table

ao508:
These changes make a couple assumptions:

1. The delivery date referenced for calculating the embargo date will always be available (never null)
2. The delivery date parsed will always be valid and can be parsed
3. We can always rely on using a substring of the calculated embargo date ISO string to get the YYYY-MM-DD format. 

~~**Why not use an actual date formatter?**~~
~~I didn't want to introduce yet another new package/dependency to the project.~~

Switched to using an actual date formatter

### DIFF
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -23,6 +23,7 @@ import CheckIcon from "@material-ui/icons/Check";
 import { StatusTooltip } from "./components/StatusToolTip";
 import { parseUserSearchVal } from "../utils/parseSearchQueries";
 import { Dispatch, SetStateAction } from "react";
+import moment from "moment";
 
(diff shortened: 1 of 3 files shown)


frontend/src/shared/helpers.tsx
```diff
@@ -990,6 +995,9 @@ export function prepareSampleCohortDataForAgGrid(samples: Sample[]) {
       });
     });
     const deliveryDate = cohortDates?.sort()[0]; // earliest cohort date
+    var embargoDateAsDate = new Date(deliveryDate);
```
alisman, line 998: @ao508 i like to use a library called Moment.js https://momentjs.com/ for this kind of date handling stuff.  But up to you.

alisman (approved):

PR #126, opened 2024-05-06
Generalize web cert properties

ao508:


qu8n (approved):
This looks great. Could I get your help removing the blurb in the README on SSL certs starting at [line 32](https://github.com/mskcc/smile-dashboard/blob/master/README.md?plain=1#L32)?

PR #128, opened 2024-05-14
Fix the disabled state of Submit Updates when editing Cost Center

qu8n:
After correcting a mistake when editing the Cost Center field, the "Submit Updates" button remains greyed out when it should be re-enabled.

Before:

https://github.com/mskcc/smile-dashboard/assets/86090707/f95916ba-fbdb-4378-b327-ad9910ea29a8

After:

https://github.com/mskcc/smile-dashboard/assets/86090707/745c3504-053d-435d-a8d4-c6a91d13b32a


frontend/src/components/SamplesList.tsx
```diff
@@ -151,10 +151,8 @@ export default function SamplesList({
         setShowAlertModal(true);
       } else {
         const allRowNodesInView = rowNode.parent?.allLeafChildren;
-        const allRowsHaveValidCostCenter = allRowNodesInView?.every(
-          (rowNode) =>
-            rowNode.data?.costCenter &&
```
qu8n, line 156: `rowNode.data?.costCenter &&` would return `false` when running into empty values. Removed it because we do want to consider empty values in `costCenter` to be valid.

frontend/src/components/SamplesList.tsx
```diff
@@ -131,6 +131,13 @@ export default function SamplesList({
     const fieldName = params.colDef.field!;
     const { oldValue, newValue, node: rowNode } = params;
 
+    function resetAlertIfCostCentersAreAllValid(changes: SampleChange[]) {
+      const allRowsHaveValidCostCenter = changes.every(
+        (c) => c.fieldName !== "costCenter" || isValidCostCenter(c.newValue)
+      );
+      if (allRowsHaveValidCostCenter) setAlertContent(defaultAlertContent);
+    }
```
qu8n, line 139: We're using the alert message state as an indicator for whether to disable the "Submit Updates" button. I defined this function here and am calling it in places where we can enable it again below.

frontend/src/components/SamplesList.tsx
```diff
@@ -141,25 +148,10 @@ export default function SamplesList({
...
-      } else {
-        const allRowNodesInView = rowNode.parent?.allLeafChildren;
-        const allRowsHaveValidCostCenter = allRowNodesInView?.every(
-          (rowNode) =>
-            rowNode.data?.costCenter &&
-            isValidCostCenter(rowNode.data.costCenter)
-        );
-        if (allRowsHaveValidCostCenter) setAlertContent(defaultAlertContent);
-      }
-    }
```
qu8n, line 161: Moved this block of code down below, to be after where we update the `changes` state array, because it relies on this `changes` state array to check whether all the edited Cost Center fields have a valid format.

PR #130, opened 2024-05-17
Surface additional fields in Cohort Samples table

ao508:
These changes surface additional fields in the Cohort Samples table.

Fields surfaced:
- custodianInformation (as Data Custodian)
- accessLevel (as Access Level)

These changes also included related updates to the graphql data schema 

image

### DIFF
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -722,13 +722,20 @@ export const CohortSampleDetailsColumns: ColDef[] = [
   {
     field: "costCenter",
     headerName: "Cost Center/Fund Number",
-    editable: true,
   },
   {
     field: "billedBy",
     headerName: "Edited By",
     headerTooltip: 'User who last updated the "Billed" status',
     headerComponentParams: createCustomHeader(lockIcon + toolTipIcon),
+  },
+  {
+    field: "custodianInformation",
+    headerName: "Data Custodian",
+  },
+  {
+    field: "accessLevel",
+    headerName: "Access Level",
   },
   {
     field: "bamCompleteDate",
@@ -792,6 +799,8 @@ const editableSampleFields = [
   "sex",
   "billed",
   "costCenter",
+  "custodianInformation",
+  "accessLevel",
 ];
 
 export function sampleFilterWhereVariables(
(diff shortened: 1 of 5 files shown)


frontend/src/shared/helpers.tsx
```diff
@@ -730,6 +730,16 @@ export const CohortSampleDetailsColumns: ColDef[] = [
...
   },
+  {
+    field: "custodianInformation",
+    headerName: "Data Custodian",
+    editable: true,
+  },
+  {
+    field: "accessLevel",
+    headerName: "Access Level",
+    editable: true,
```
qu8n, line 741: In our current setup, editability is enabled by adding the editable field names to the `editableSampleFields` array (currently at [line 791](https://github.com/mskcc/smile-dashboard/pull/130/files#diff-db966dd79103f32fddb49dcefcfba78d4325506ce364cd0e24779a3521aa7173R791)) instead of enabling it at the individual column definition level like here. It appears these fields are not editable right now in dev, at least on my end.

For next step, we should be good to go after removing `editable: true` here and adding `custodianInformation` and `accessLevel` to the `editableSampleFields` array.

(For context, this configuration currently happens at [lines 510-515](https://github.com/mskcc/smile-dashboard/pull/130/files#diff-db966dd79103f32fddb49dcefcfba78d4325506ce364cd0e24779a3521aa7173R510-R515). We do this to [1] save ourselves from manually adding `editable: true/false` to every column definition and [2] make clearer at a quick glance which fields we allow users to edit.)

qu8n (changes requested):
Looks great! All is good on my end after this one quick change request. Thank you.

**Edit:**
The notes in [ticket #1125](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1125) mentioned that Access Level value is based on the embargo date. Should/do we have logic somewhere to reflect that? Or will this be manually updated by the PMs? 

Looks like we do set the default value for Access Level if it's empty in smile-server [here](https://github.com/mskcc/smile-server/commit/7a9cd34b4e197a127bf9802538311dd9ed84c184) but that's all I could find.

qu8n (approved):

PR #132, opened 2024-05-23
Fix filtered view's row count

qu8n:


frontend/src/components/RecordsList.tsx
qu8n, line 1: Renamed `remoteCount` in this `RecordsList` component to match the same variable in the `SamplesList` component. Also not sure why this was called "remoteCount" in the first place and thought that "rowCount" was clearer.

PR #133, opened 2024-05-23
Init commit - search cohorts by sample id

ao508:


frontend/src/components/RecordsList.tsx
```diff
@@ -113,6 +113,11 @@ export default function RecordsList({
           [`${dataName}ConnectionWhere2`]: {
             OR: queryFilterWhereVariables(parsedSearchVals),
           },
+          hasCohortSampleSamplesWhere2: {
+            hasMetadataSampleMetadata_SOME: {
+              primaryId_IN: parsedSearchVals,
+            },
+          },
```
ao508, line 120: This needs to be generalized

ao508 (commented):
Some notes:


Sample `10786_D_433`

Cohorts:
- CCS_SA7WUHQB
- CCS_AVGQFYAA


- Filters results to only 2 cohorts (above sample is in fact in 2 cohorts)
- the cohorts' 'hasCohortSampleSamplesConnection' total count = the total number of samples for that cohort 


- **hasCohortSampleSamplesWhere2** - filters results to the two cohorts containing the sample primary id being searched for
- **cohortsConnectionWhere2** - makes it so that the total count matches the total number of cohorts returned by the query 
- **where** - removing this returns all cohorts instead of just the cohorts matching the sample primary id being searched for
```
{
  "hasCohortSampleSamplesWhere2": {
    "hasMetadataSampleMetadata_SOME": {
      "primaryId_IN": ["10786_D_433"]
    }
  },
  "cohortsConnectionWhere2": {
    "hasCohortSampleSamples_SOME": {
      "hasMetadataSampleMetadata_SOME": {
        "primaryId_IN": ["10786_D_433"]
      }
    }
  },
  "where": {
    "hasCohortSampleSamples_SOME": {
      "hasMetadataSampleMetadata_SOME": {
        "primaryId_IN": ["10786_D_433"]
      }
    }
  }
}
```

frontend/src/components/RecordsList.tsx
ao508: The other tables still render fine without having to generalize this specific "where" variable but for consistency we should probably move forward with generalizing it the way that the other query variables are generalized

PR #134, opened 2024-05-30
Fix filtered view count and enable cohort filter by delivery date

qu8n:
This PR includes the work for both tickets [#1154](https://github.com/mskcc/smile-server/issues/1154) and [#1131](https://github.com/mskcc/smile-server/issues/1131).

https://github.com/mskcc/smile-dashboard/assets/86090707/6690069e-ed59-4f4b-aea1-7733b90d5244

### DIFF
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -46,8 +46,6 @@ interface IRecordsListProps {
   showDownloadModal: boolean;
   setShowDownloadModal: Dispatch<SetStateAction<boolean>>;
   handleDownload: () => void;
-  setFilterModel?: Dispatch<SetStateAction<Record<string, any>>>;
-  customFilterWhereVariables?: Record<string, any>;
   samplesQueryParam: string | undefined;
   prepareSamplesDataForAgGrid?: (samples: Sample[]) => any[];
   samplesColDefs: ColDef[];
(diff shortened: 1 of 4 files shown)


frontend/src/components/RecordsList.tsx
```diff
@@ -29,6 +29,7 @@ import { ErrorMessage, LoadingSpinner, Toolbar } from "../shared/tableElements";
 interface IRecordsListProps {
   colDefs: ColDef[];
   dataName: DataName;
+  enableInfiniteScroll?: boolean;
```
qu8n, line 32: As a refresher, `RecordsList` is the component that renders Requests, Patients, and Cohorts pages. By default, this will be on. For the Cohorts page, we will turn this off to enable client-side filtering.

This will make the Cohorts page fetch the entire dataset to the browser on each render, but I'm not concerned about performance because the Cohorts dataset is tiny.

frontend/src/shared/helpers.tsx
```diff
@@ -577,25 +580,44 @@ export function prepareCohortDataForAgGrid(
...
       newCohortsConnection.totalCount = 0;
     } else {
-      newCohorts = newCohorts.filter(
-        (cohort) => cohort.billed === selectedValues[0]
+      newCohorts = newCohorts.filter((cohort) =>
+        selectedValues.includes(cohort.billed)
       );
       newCohortsConnection.totalCount = newCohorts.length;
     }
   }
```
qu8n, line 608: Every time the user performs a filter, the entire Cohorts dataset is fetched, then this code filters the data based on the user's filter inputs, and finally a filtered data table is rendered to the user.

ao508 (approved):

PR #137, opened 2024-06-11
Add Tempo fields to the Samples page

qu8n:
This PR adds additional Tempo-related fields to the table on Samples page. Note that while these Tempo fields are editable on the Cohort Sample view, these fields are read-only on the Samples page.

https://github.com/mskcc/smile-dashboard/assets/86090707/12306735-c104-4113-bcfb-54255c059b0a

### DIFF
--- a/frontend/src/pages/patients/PatientsPage.tsx
+++ b/frontend/src/pages/patients/PatientsPage.tsx
@@ -15,7 +15,7 @@ import InfoIcon from "@material-ui/icons/InfoOutlined";
 import { parseUserSearchVal } from "../../utils/parseSearchQueries";
 import {
   PatientsListColumns,
-  SampleDetailsColumns,
+  SampleMetadataDetailsColumns,
   preparePatientDataForAgGrid,
   sampleFilterWhereVariables,
 } from "../../shared/helpers";
@@ -269,7 +269,7 @@ export default function PatientsPage({
           });
           setShowDownloadModal(true);
         }}
-        samplesColDefs={SampleDetailsColumns}
+        samplesColDefs={SampleMetadataDetailsColumns}
         samplesQueryParam={sampleQueryParamValue && "Patient's Samples"}
         samplesParentWhereVariables={
           {
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -911,6 +911,12 @@ export function cohortSampleFilterWhereVariables(
         costCenter_IN: parsedSearchVals,
       },
       {
+        custodianInformation_IN: parsedSearchVals,
+      },
+      {
+        accessLevel_IN: parsedSearchVals,
+      },
+      {
         hasEventBamCompletes_SOME: {
           date_IN: parsedSearchVals,
         },
@@ -963,6 +969,12 @@ export function cohortSampleFilterWhereVariables(
       },
       {
         costCenter_CONTAINS: parsedSearchVals[0],
+      },
+      {
+        custodianInformation_CONTAINS: parsedSearchVals[0],
+      },
+      {
+        accessLevel_CONTAINS: parsedSearchVals[0],
       },
       {
         hasEventBamCompletes_SOME: {
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -723,7 +723,7 @@ export const CohortSampleDetailsColumns: ColDef[] = [
     headerName: "CMO Sample Name",
   },
   {
-    field: "deliveryDate",
+    field: "initialPipelineRunDate",
     headerName: "Initial Pipeline Run Date",
   },
   {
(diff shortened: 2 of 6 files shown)


frontend/src/pages/patients/PatientsPage.tsx
```diff
@@ -15,7 +15,7 @@ import InfoIcon from "@material-ui/icons/InfoOutlined";
 import { parseUserSearchVal } from "../../utils/parseSearchQueries";
 import {
   PatientsListColumns,
-  SampleDetailsColumns,
+  SampleMetadataDetailsColumns,
```
qu8n, line 18: Just a rename. We have a `CohortSampleDetailsColumns` so I wanted to specify that this variable is for SampleMetadata.

frontend/src/shared/helpers.tsx
```diff
@@ -475,13 +476,15 @@ function setupEditableSampleFields(samplesColDefs: ColDef[]) {
         });
         return changedValue !== undefined;
       },
-      "costCenter-validation-error": (params: CellClassParams) => {
-        if (params.colDef.field === "costCenter") {
-          return !isValidCostCenter(params.value);
-        }
-        return false;
-      },
```
qu8n, line 483: Moved this CSS class setup from inside this function to `CohortSampleDetailsColumns`. The reason is to keep this function strictly for setting up editable fields (per its name `setupEditableSampleFields`).

frontend/src/shared/helpers.tsx
```diff
@@ -804,9 +815,25 @@ export const CohortSampleDetailsColumns: ColDef[] = [
...
+  ...SampleMetadataDetailsColumns,
+  ...ReadOnlyCohortSampleDetailsColumns,
+].filter((col) => {
+  if (seenColumns.has(col.field)) {
+    return false;
+  }
+  seenColumns.add(col.field);
+  return true;
+});
+
```
qu8n, line 836: Created a read-only copy of `CohortSampleDetailsColumns`, then combined that with `SampleMetadataDetailsColumns` to form a variable with *unique* column definitions.

frontend/src/shared/helpers.tsx
```diff
@@ -952,6 +985,12 @@ export function cohortSampleFilterWhereVariables(
       {
         costCenter_CONTAINS: parsedSearchVals[0],
       },
+      {
+        custodianInformation_CONTAINS: parsedSearchVals[0],
+      },
+      {
+        accessLevel_CONTAINS: parsedSearchVals[0],
+      },
```
qu8n, line 993: Realized these fields were never included in the search functionality for Tempo fields, so I added them.

frontend/src/shared/helpers.tsx
```diff
@@ -1038,72 +1077,99 @@ export function prepareSampleMetadataForAgGrid(samples: Sample[]) {
...
+
+    return {
+      revisable: s.revisable,
+      ...sampleMetadata,
+      ...tempoData,
+    };
+  });
+}
+
+function extractTempoFromSample(s: Sample) {
```
qu8n, line 1106: The code inside this new function is mostly the same code found in the previous definition of `prepareSampleCohortDataForAgGrid()`. I generalized it into this named function so that it could be used for both (1) the existing Cohort Sample view and (2) the new Samples page.

frontend/src/shared/helpers.tsx
```diff
@@ -804,9 +815,25 @@ export const CohortSampleDetailsColumns: ColDef[] = [
...
 
-setupEditableSampleFields(SampleDetailsColumns);
+export const ReadOnlyCohortSampleDetailsColumns = _.cloneDeep(
+  CohortSampleDetailsColumns
+);
+
+setupEditableSampleFields(SampleMetadataDetailsColumns);
 setupEditableSampleFields(CohortSampleDetailsColumns);
 
+const seenColumns = new Set();
```
alisman, line 825: this is kind of odd.  you should put this set up in some constructor.  also, you have a classic "side effect" here.  you are populating seenColumns inside of the filter function, which, per functional programming paradigm, should not mutate anything outside.  it should just return a value.    also, i think that the Set will take care of uniqueness by itself.  double check this but i think a Set will not allow duplicate values.

frontend/src/shared/helpers.tsx
```diff
@@ -1038,72 +1077,99 @@ export function prepareSampleMetadataForAgGrid(samples: Sample[]) {
...
+    embargoDateAsDate.setMonth(embargoDateAsDate.getMonth() + 18);
+    embargoDate = moment(embargoDateAsDate).format("YYYY-MM-DD");
+  }
 
-    const tempo = s.hasTempoTempos?.[0];
-    const { billedBy, costCenter, custodianInformation, accessLevel } =
-      tempo ?? {};
+  const tempo = s.hasTempoTempos?.[0];
+  const { billedBy, costCenter, custodianInformation, accessLevel } =
+    tempo ?? {};
```
alisman, line 1125: wow. fancy.  are you sure that it's not just ```tempo || {}``` ?

qu8n: I think both could work here. Looks like `??` lets us return `{}` if `tempo` is `undefined` or `null`, whereas `||` returns `{}` if `tempo` is anything falsy (broader cases).

I think `const tempo = s.hasTempoTempos?.[0];` could only either return some defined value or `undefined`, so I think `??` should be ok. 

Or should we use `||` to be safe?

frontend/src/shared/helpers.tsx
qu8n: @alisman Thanks, Aaron. Good call on the functional programming paradigm. 

Thoughts on the following code as replacement? All in one line so nothing outside is mutated. I confirmed that this achieves the same result.
```ts
export const combinedSampleDetailsColumns = 
    [...SampleMetadataDetailsColumns, ...ReadOnlyCohortSampleDetailsColumns]
        .filter((col, index, self) => index === 
            self.findIndex((c) => c.field === col.field));
```

Unfortunately, I couldn't use Set here because of 2 reasons:
1. The order of the columns in the data table is important
2. Set doesn't remove dups from an array of objects (even if the content of objects match completely) as each object has a unique reference

alisman: i still find it very difficult to understand that code!  you're filtering for items (col) where the index is equal to the index of an item where field matches ...  ??   
you are just trying to remove duplicate objects which are not referentially equal?
if so, you could use _.uniqBy.    this would result in something much more readable.

frontend/src/shared/helpers.tsx
alisman: ?? is fine.  i just hadn't seen it before.

frontend/src/shared/helpers.tsx
```diff
@@ -1038,72 +1077,99 @@ export function prepareSampleMetadataForAgGrid(samples: Sample[]) {
...
+  let billed = tempo?.billed;
+  if (!billed) {
+    billed = false;
+  }
 
-    const bamComplete = tempo?.hasEventBamCompletes?.[0];
-    const { date: bamCompleteDate, status: bamCompleteStatus } =
-      bamComplete ?? {};
+  const bamComplete = tempo?.hasEventBamCompletes?.[0];
+  let { date: bamCompleteDate, status: bamCompleteStatus } = bamComplete ?? {};
```
alisman, line 1135: looking at this destructuring here, i'm not sure what it accomplishes.  in other words, where you use these values, why not just use dot notation on the source object.    like, bamComplete.date strikes me as actually better than assigning to bamCompleteDate because then the reader doesn't have to figure out where it's assigned.  sorry if i'm missing something.

qu8n: @alisman We could use individual assignments instead of object destructuring, but ultimately we'll need some unique variable name (`bamCompleteDate` in case) as a field name for the AG Grid column definitions to reference:
```ts
export const CohortSampleDetailsColumns: ColDef[] =
[... 
{
    field: "bamCompleteDate",
    headerName: "Latest BAM Complete Date",
}, 
...]
```

I could update the code like this. Would you prefer this / find it to be more readable?


<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
  
```ts
  const tempo = s.hasTempoTempos?.[0];
  const { billedBy, costCenter, custodianInformation, accessLevel } =
    tempo ?? {};

  const bamComplete = tempo?.hasEventBamCompletes?.[0];
  let { date: bamCompleteDate, status: bamCompleteStatus } = bamComplete ?? {};
  bamCompleteDate = formatCohortRelatedDate(bamCompleteDate);

  const mafComplete = tempo?.hasEventMafCompletes?.[0];
  let {
    date: mafCompleteDate,
    status: mafCompleteStatus,
    normalPrimaryId: mafCompleteNormalPrimaryId,
  } = mafComplete ?? {};
  mafCompleteDate = formatCohortRelatedDate(mafCompleteDate);

  const qcComplete = tempo?.hasEventQcCompletes?.[0];
  let {
    date: qcCompleteDate,
    result: qcCompleteResult,
    reason: qcCompleteReason,
    status: qcCompleteStatus,
  } = qcComplete ?? {};
  qcCompleteDate = formatCohortRelatedDate(qcCompleteDate);
```
  
</td>
<td>

```ts
  const tempo = s.hasTempoTempos?.[0];
  const billedBy = tempo?.billedBy;
  const costCenter = tempo?.costCenter;
  const custodianInformation = tempo?.custodianInformation;
  const accessLevel = tempo?.accessLevel;

  const bamComplete = tempo?.hasEventBamCompletes?.[0];
  const bamCompleteDate = formatCohortRelatedDate(bamComplete?.date);
  const bamCompleteStatus = bamComplete?.status;

  const mafComplete = tempo?.hasEventMafCompletes?.[0];
  const mafCompleteDate = formatCohortRelatedDate(mafComplete?.date);
  const mafCompleteStatus = mafComplete?.status;
  const mafCompleteNormalPrimaryId = mafComplete?.normalPrimaryId;
  
  const qcComplete = tempo?.hasEventQcCompletes?.[0];
  const qcCompleteDate = formatCohortRelatedDate(qcComplete?.date);
  const qcCompleteResult = qcComplete?.result;
  const qcCompleteReason = qcComplete?.reason;
  const qcCompleteStatus = qcComplete?.status;
```

</td>
</tr>
</table>

frontend/src/shared/helpers.tsx
qu8n: This is way cleaner and works. Awesome.

```ts
export const combinedSampleDetailsColumns = _.uniqBy(
  [...SampleMetadataDetailsColumns, ...ReadOnlyCohortSampleDetailsColumns],
  'field'
);
```

frontend/src/shared/helpers.tsx
alisman: it is more readable, but a further question is, why are you bothering to assign to local variables here when you are just returning the values below?  i.e. why not just start at line 1155 and return soemthing like 
```
return {
    billed:mafComplete?.whatever
....
}
```

alisman (approved):

ao508 (approved):

PR #138, opened 2024-06-18
Fix data export across all views

qu8n:
This PR resolves the following bugs:
- Missing data in the export file of the Patients and Cohorts pages
- Missing MRN in the export file of the Patients page
- PHI warning appearing even when exporting in non-PHI mode on Patients page


ao508 (approved):
Approved! 

I did notice that the exported patient table doesn't follow the same filename format that we use for the other tables but I think this is an old bug so I'm making a separate ticket for it [here](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1193)

PR #140, opened 2024-07-01
Flatten data queried to SSRM AG Grid tables

qu8n:
For [card #1211](https://github.com/mskcc/smile-server/issues/1211).

This PR "flattens" the GraphQL-queried data from Neo4j, bringing nested fields to the root level.
<table>
<tr>
<th> Before </th>
<th> After </th>
</tr>
<tr>
<td>

```gql
node {
  field1
  hasChildNode {
    field2
    field3
  }
}
```

</td>
<td>

```gql
node {        
  field1        
  field2        
  field3        
}



```

</td>
</tr>
</table>

This "flattening" is done by adding these fields to the Neo4j schema, then writing resolvers so that the Apollo Client knows how to return data for these fields.

Pros:
 - Right now, we’re unable to perform sorting of nested fields as that isn't natively supported by GraphQL[^1]. Having custom resolvers for these nested fields allow us to write logic to handle sorting in a future ticket.
 - Moving the data processing logic to the GraphQL/Apollo server and leaving the frontend responsible for just the “view” (better separation of concerns).
 - Ease of access to nested fields for the frontend.

Cons:
 - Generally, having fewer custom resolvers is better. Unfortunately, there is no workarounds at the moment to handle sorting of nested fields.
 - It's less clear which types in the "Neo4j" schema were generated directly from the Neo4j database, and which are the custom types. To handle this, we'll need to be more diligent with documentations (through self-documenting code and the README) to ensure future contributors understand the what and why of the extended types.

[^1]: GraphQL fields are resolved in a top-down fashion. That means the root-level fields are resolved first, then then fields in the next nested level, and so on.

### DIFF
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -32,7 +32,7 @@ interface IRecordsListProps {
   enableInfiniteScroll?: boolean;
   lazyRecordsQuery: typeof useHookLazyGeneric;
   lazyRecordsQueryAddlVariables?: Record<string, any>;
-  prepareDataForAgGrid: (
+  prepareDataForAgGrid?: (
     data: any,
     filterModel: IServerSideGetRowsRequest["filterModel"]
   ) => any;
@@ -135,7 +135,10 @@ export default function RecordsList({
               });
 
         return thisFetch.then((d) => {
-          const agGridData = prepareDataForAgGrid(d.data, filterModel);
+          let agGridData = d.data;
+          if (prepareDataForAgGrid) {
+            agGridData = prepareDataForAgGrid(d.data, filterModel);
+          }
 
           if ("uniqueSampleCount" in agGridData) {
             setUniqueSampleCount(agGridData.uniqueSampleCount);
@@ -180,7 +183,10 @@ export default function RecordsList({
                 },
               },
             }).then(({ data }) => {
-              const agGridData = prepareDataForAgGrid(data, filterModel);
+              let agGridData = data;
+              if (prepareDataForAgGrid) {
+                agGridData = prepareDataForAgGrid(data, filterModel);
+              }
               return buildTsvString(agGridData[dataName], colDefs);
             });
           }}
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -116,9 +116,6 @@ export default function RecordsList({
           where: {
             OR: queryFilterWhereVariables(parsedSearchVals),
           },
-          [`${dataName}ConnectionWhere2`]: {
-            OR: queryFilterWhereVariables(parsedSearchVals),
-          },
           options: {
             offset: params.request.startRow,
             limit: params.request.endRow,
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -552,9 +552,7 @@ export function prepareCohortDataForAgGrid(
       totalSamples: cohort.hasCohortSampleSamplesConnection.totalCount,
       smileSampleIds: samples.map((s) => s.smileSampleId),
       billed: allSamplesBilled === true ? "Yes" : "No",
-      initialCohortDeliveryDate: formatDate(
-        cohortCompletes?.slice(-1)[0]?.date
-      ),
+      initialCohortDeliveryDate: formatDate(cohort.initialCohortDeliveryDate),
       completeDate: formatDate(latestCohortDeliveryDate?.date),
       endUsers: latestCohortDeliveryDate?.endUsers,
       pmUsers: latestCohortDeliveryDate?.pmUsers,
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -13,7 +13,6 @@ import "ag-grid-enterprise";
 import {
   CohortsListQuery,
   PatientsListQuery,
-  RequestsListQuery,
   Sample,
   SampleMetadata,
   SampleMetadataWhere,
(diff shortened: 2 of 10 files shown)


frontend/src/components/RecordsList.tsx
```diff
@@ -32,7 +32,7 @@ interface IRecordsListProps {
   enableInfiniteScroll?: boolean;
   lazyRecordsQuery: typeof useHookLazyGeneric;
   lazyRecordsQueryAddlVariables?: Record<string, any>;
-  prepareDataForAgGrid: (
+  prepareDataForAgGrid?: (
```
qu8n, line 35: This is the function that previously flattens the data for AG Grid to use. Now that data is flattened before reaching the frontend, this prop is now optional. We can't remove it because the Cohorts page uses this to filter the data (see `function prepareCohortDataForAgGrid` in `helpers.tsx`).

frontend/src/shared/helpers.tsx
```diff
@@ -145,64 +143,6 @@ export const RequestsListColumns: ColDef[] = [
...
-      consentPartC: additionalPropertiesJson["consent-partc"],
-    };
-  });
-
-  return {
-    patientsConnection: patientsListQueryResult.patientsConnection,
-    patients: newPatients,
-  };
-}
-
```
qu8n, line 205: Removed functions that flatten data on the frontend. This step is now done at the GraphQL server level.

frontend/src/shared/helpers.tsx
```diff
@@ -535,36 +475,7 @@ export function prepareCohortDataForAgGrid(
...
-      completeDate: formatDate(latestCohortDeliveryDate?.date),
-      endUsers: latestCohortDeliveryDate?.endUsers,
-      pmUsers: latestCohortDeliveryDate?.pmUsers,
-      projectTitle: latestCohortDeliveryDate?.projectTitle,
-      projectSubtitle: latestCohortDeliveryDate?.projectSubtitle,
-      status: latestCohortDeliveryDate?.status,
-      type: latestCohortDeliveryDate?.type,
-    };
-  });
-
```
qu8n, line 567: Removed code that flatten data on the frontend. This step is now done at the GraphQL server level.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -21,11 +24,13 @@ import { connect, headers, StringCodec } from "nats";
...
 const request = require("request-promise-native");
 import { ApolloClient, ApolloQueryResult } from "apollo-client";
+import { gql } from "apollo-server";
 
 export async function buildNeo4jDbSchema() {
   const driver = neo4j.driver(
     props.neo4j_graphql_uri,
-    neo4j.auth.basic(props.neo4j_username, props.neo4j_password)
+    neo4j.auth.basic(props.neo4j_username, props.neo4j_password),
+    { disableLosslessIntegers: true } // maps Cypher Integer to JavaScript Number
```
qu8n, line 33: This allowed us to use Cypher type `Int` when extending the type definitions below. 

More context directly from the docs:
> [Integer](https://neo4j.com/docs/api/javascript-driver/current/class/lib6/integer.js~Integer.html) is not one of JavaScript’s native types, but rather a custom one accomodating Cypher’s precision. You can disable this through the [disableLosslessIntegers configuration entry](https://neo4j.com/docs/api/javascript-driver/current/function/index.html#configuration) when instantiating the driver, so that JavaScript’s native Number type is used instead. Note that this can lead to a loss of precision.

(I think loss of precision is fine as we only use integers for our application.)

graphql-server/src/schemas/neo4j.ts
```diff
@@ -42,9 +47,39 @@ export async function buildNeo4jDbSchema() {
...
+      billed: String
+      initialCohortDeliveryDate: String
+      endUsers: String
+      pmUsers: String
+      projectTitle: String
+      projectSubtitle: String
+      status: String
+      type: String
+    }
+  `;
```
qu8n, line 78: Added extra fields to the schema. This let us bring these nested fields to the root level of a node.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -142,6 +177,90 @@ function buildResolvers(
...
+      projectSubtitle: (parent: CohortsListQuery["cohorts"][number]) => {
+        return parent.hasCohortCompleteCohortCompletes?.[0]?.projectSubtitle;
+      },
+      status: (parent: CohortsListQuery["cohorts"][number]) => {
+        return parent.hasCohortCompleteCohortCompletes?.[0]?.status;
+      },
+      type: (parent: CohortsListQuery["cohorts"][number]) => {
+        return parent.hasCohortCompleteCohortCompletes?.[0]?.type;
+      },
+    },
```
qu8n, line 263: Custom resolvers for the newly added fields in the schema.

graphql/operations.graphql
qu8n, line 1: Removed the `$requestsConnectionWhere2`, `$patientsConnectionWhere2`, and `$cohortsConnectionWhere2` variables from the queries below because they're redundant, and replaced each with the `$where` variable as they both receive the same input (see [here](https://github.com/mskcc/smile-dashboard/blob/master/frontend/src/components/RecordsList.tsx#L116-L121)).

ao508 (approved):
Approved. There's an issue with the filenames generated on the Cohort Samples, Request Samples,. and Patient Samples (pt table already known) but I don't think that that should prevent these changes from getting merged. 

Made a new issue for the filename bug [here](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1222)

PR #141, opened 2024-07-08
Auto sort the Requests and Cohorts views

qu8n:
This PR is a superset of #140, so we can either merge them both in order or just merge this one.

This PR introduces the following changes:
- Tab name updates in Samples page
- Show Import Date in Requests view
- Auto-sort the initial load of Requests view by Import Date
- Auto-sort the initial load of Cohorts view by Initial Cohort Delivery Date

Note that this PR only applies sorting to the initial load of the aforementioned views. The ability to manually sort fields will be introduced in a future PR for card [#1152](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1152).

### DIFF
--- a/graphql-server/src/schemas/neo4j.ts
+++ b/graphql-server/src/schemas/neo4j.ts
@@ -178,6 +178,59 @@ function buildResolvers(
         };
       },
     },
+    Query: {
+      async requests(_source: undefined, args: any) {
+        const requests = await ogm.model("Request").find({
+          where: args.where,
+          options: args.options,
+          selectionSet: `{
+            igoRequestId
+            igoProjectId
+            genePanel
+            dataAnalystName
+            dataAnalystEmail
+            dataAccessEmails
+            bicAnalysis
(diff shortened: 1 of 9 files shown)


graphql-server/src/schemas/neo4j.ts
```diff
@@ -142,6 +178,196 @@ function buildResolvers(
...
+                return importDateA > importDateB ? 1 : -1;
+              } else {
+                return importDateA < importDateB ? 1 : -1;
+              }
+            });
+          }
+        }
+
+        return requests;
+      },
```
qu8n, line 232: This code handles the initial sorting in the Requests view. It works by first (1) querying the Requests data from Neo4j using the OGM with the same format as `query RequestsList` and (2) apply the sorting logic.

A similar flow applies to the `cohorts` query resolver below.

ao508 (approved):

PR #143, opened 2024-07-18
Enable sorting for all columns in Requests, Patients, and Cohorts views

qu8n:
For card [#1152](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1152).

This PR enables users to sort any of the fields in Requests, Patients, and Cohorts views, including nested fields.

Demo on sorting Request columns:

https://github.com/user-attachments/assets/c0986e09-0c08-47df-8329-d42f88c4f4b6

### DIFF
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -1,6 +1,6 @@
 import AutoSizer from "react-virtualized-auto-sizer";
 import { Button, Container, Modal } from "react-bootstrap";
-import { Dispatch, SetStateAction, useMemo, useRef } from "react";
+import { Dispatch, SetStateAction, useCallback, useMemo, useRef } from "react";
 import { useNavigate } from "react-router-dom";
 import { DownloadModal } from "./DownloadModal";
 import { buildTsvString } from "../utils/buildTsvString";
@@ -117,7 +117,7 @@ export default function RecordsList({
 
   const totalCountNodeName = `${dataName}Connection`;
 
-  function getSortModel() {
+  const getSortModel = useCallback(() => {
     const sortModel = grid.current?.columnApi
       ?.getColumnState()
       .filter(({ sort }) => sort)
@@ -128,7 +128,7 @@ export default function RecordsList({
       return sortModel;
     }
     return defaultSort;
-  }
+  }, [defaultSort]);
 
   const datasource: IServerSideDatasource = useMemo(() => {
     return {
--- a/frontend/src/components/RecordsList.tsx
+++ b/frontend/src/components/RecordsList.tsx
@@ -1,6 +1,6 @@
 import AutoSizer from "react-virtualized-auto-sizer";
 import { Button, Container, Modal } from "react-bootstrap";
-import { Dispatch, SetStateAction, useMemo } from "react";
+import { Dispatch, SetStateAction, useMemo, useRef } from "react";
 import { useNavigate } from "react-router-dom";
 import { DownloadModal } from "./DownloadModal";
 import { buildTsvString } from "../utils/buildTsvString";
@@ -25,6 +25,7 @@ import {
 } from "../shared/helpers";
 import { PatientIdsTriplet } from "../pages/patients/PatientsPage";
 import { ErrorMessage, LoadingSpinner, Toolbar } from "../shared/tableElements";
+import { AgGridReact as AgGridReactType } from "ag-grid-react/lib/agGridReact";
 
 interface IRecordsListProps {
   colDefs: ColDef[];
@@ -96,6 +97,7 @@ export default function RecordsList({
   const [rowCount, setRowCount] = useState<number>(0);
   const [uniqueSampleCount, setUniqueSampleCount] = useState<number>(0);
 
+  const grid = useRef<AgGridReactType>(null);
   const navigate = useNavigate();
 
   // note that we aren't using initial fetch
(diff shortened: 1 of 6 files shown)


frontend/src/components/RecordsList.tsx
```diff
@@ -94,20 +98,38 @@ export default function RecordsList({
...
+      ?.getColumnState()
+      .filter(({ sort }) => sort)
+      .map(({ colId, sort }) => ({
+        [colId]: sort?.toUpperCase() as SortDirection,
+      }));
+    if (sortModel && sortModel.length > 0) {
+      return sortModel;
+    }
+    return defaultSort;
+  }, [defaultSort]);
```
qu8n, line 131: This function gets called whenever the frontend queries. It helps the frontend pass along the sort setting to the GraphQL custom resolver. If the user manually sort some column, use that. Otherwise, use the default sort setting (e.g. `importDate` for Requests view).

frontend/src/shared/helpers.tsx
```diff
@@ -512,6 +510,33 @@ export function prepareCohortDataForAgGrid(
...
+        return sortOrder === "ASC" ? a - b : b - a;
+      } else if (typeof a === "string" && typeof b === "string") {
+        return sortOrder === "ASC"
+          ? a.localeCompare(b, "en", { sensitivity: "base" })
+          : b.localeCompare(a, "en", { sensitivity: "base" });
+      } else {
+        return 0;
+      }
+    });
+  }
```
qu8n, line 538: Handle sorting for Cohorts view where we're filtering and now sorting from the frontend. Initially, I planned on moving all of this logic of the Cohorts view to the GraphQL server, but I'll save that work for [this card](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1229).

graphql-server/src/schemas/neo4j.ts
```diff
@@ -177,93 +182,326 @@ function buildResolvers(
...
+
+          if (["importDate", "totalSampleCount"].includes(sortField)) {
+            sortArrayByNestedField(requests, "Request", sortField, sortOrder);
+          }
+        }
+
+        if (args.options?.limit == null) {
+          return requests;
+        }
+        return requests.slice(args.options.offset, args.options.limit);
```
qu8n, line 233: After the frontend makes the `requests` query, this GraphQL custom resolver gets called to handle that query. How it works:
- Use the Neo4j OGM to query the database. The `where` variable is passed along to handle searching. The `sort` is also passed in, but it will only handle sorting of non-nested fields.
- If the sort setting includes a nested field, we'll call `sortArrayByNestedField` to handle that.
- Finally, we'll return paginated data to handle infinite scroll. We'll only return non-paginated data when the user exports.

A similar flow applies to the `patients` query below.

The `cohorts` query follows a lite version of this flow because it currently handles sorting/filtering on the frontend (doesn't make additional requests when sorting/filtering).

graphql-server/src/schemas/neo4j.ts
```diff
@@ -182,7 +186,9 @@ function buildResolvers(
       async requests(_source: undefined, args: any) {
```
qu8n, line 186: After the frontend makes the `requests` query, this GraphQL custom resolver gets called to handle that query. How it works:
- Use the Neo4j OGM to query the database. The `where` variable is passed along to handle searching. The `sort` is also passed in, but it will only handle sorting of non-nested fields.
- If the sort setting includes a nested field, we'll call `sortArrayByNestedField` to handle that.
- Finally, we'll return paginated data to handle infinite scroll. We'll only return non-paginated data when the user exports.

A similar flow applies to the `patients` query below.

The `cohorts` query follows a lite version of this flow because it currently handles sorting/filtering on the frontend (doesn't make additional requests when sorting/filtering).

frontend/src/components/RecordsList.tsx
```diff
@@ -175,25 +191,28 @@ export default function RecordsList({
...
-              }
-              return buildTsvString(agGridData[dataName], colDefs);
             });
+
+            let agGridData = data;
+            if (prepareDataForAgGrid) {
+              agGridData = prepareDataForAgGrid(data, filterModel, sortModel);
+            }
+
+            return buildTsvString(agGridData[dataName], colDefs);
```
ao508, line 215: tested this myself and confirmed that the records are saved to the file in the order in which they're displayed. thanks!

frontend/src/shared/helpers.tsx
```diff
@@ -177,37 +177,30 @@ export const PatientsListColumns: ColDef[] = [
...
   },
   {
     field: "consentPartC",
     headerName: "12-245 Part C",
-    sortable: false,
   },
   {
     field: "smilePatientId",
     headerName: "SMILE Patient ID",
-    hide: true,
```
ao508, line 210: this should still be hidden by default. Was this causing any issues or was this a mistake?

frontend/src/shared/helpers.tsx
```diff
@@ -516,6 +510,33 @@ export function prepareCohortDataForAgGrid(
     });
   });
 
+  if (sortModel) {
```
ao508, line 513: maybe add a comment here with a brief explanation on what this block is doing?

graphql-server/src/schemas/neo4j.ts
```diff
@@ -283,94 +331,177 @@ function buildResolvers(
...
   };
 }
 
+function sortArrayByNestedField(
+  arr: any[],
+  nodeLabel: string,
+  fieldName: string,
+  sortOrder: SortDirection
+) {
+  arr.sort((objA, objB) => {
```
ao508, line 403: Re my comment [here](https://github.com/mskcc/smile-dashboard/pull/143/files#diff-db966dd79103f32fddb49dcefcfba78d4325506ce364cd0e24779a3521aa7173R513) - maybe the logic _there_ can be moved to a function call also like how this function is set up?

ao508 (commented):
Some minor comments and general questions. Otherwise it's looking good! Will definitely be ready to merge after further discussion

frontend/src/shared/helpers.tsx
qu8n: This was a mistake. Thanks for catching!

qu8n:
> Some minor comments and general questions. Otherwise it's looking good! Will definitely be ready to merge after further discussion

@ao508 Thanks! I just pushed some changes following your suggestions

ao508 (approved):

PR #145, opened 2024-07-29
Add cached Oncotree data to Samples views

qu8n:
PR for card [#1232](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1232).

# Description
This PR adds `cancerType` and `cancerTypeDetailed` fields corresponding to `oncotreeCode` values to views with SampleMetadata data.

Data for these fields are initially fetched from the Oncotree API's candidate release upon the GraphQL server startup, and it's then cached with a duration of 1 day (matches cBioPortal's setting). When the GraphQL custom resolver encounters a new `oncotreeCode` that isn't in the cache, it will refetch from the Oncotree API and refresh the cache.

Old `oncotreeCode` values (like "HIST") with no matching data from the Oncotree API have `cancerType` and `cancerTypeDetailed` fields as "N/A", along with a tooltip describing the issue. These "N/A" values are added to cache to avoid SMILE continuously fetching the Oncotree API for non-existent codes.

CleanShot 103738

# Notes
- PMs have the ability to update the `oncotreeCode` value to see `cancerType` and `cancerTypeDetailed` fields automatically updated
- Searching `cancerType` and `cancerTypeDetailed` fields would require additional engineering work as we'd have to write a GraphQL custom resolver for the samples query. This work is captured in card [#1244](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1244)
- We're using `node-cache` instead of the built-in [GraphQL server-side cache](https://www.apollographql.com/docs/apollo-server/performance/caching/#in-your-resolvers-dynamic) for more control and flexibility over the cache (e.g. when we need to refresh the cache on the fly).

### DIFF
--- a/frontend/src/generated/graphql.ts
+++ b/frontend/src/generated/graphql.ts
@@ -7524,6 +7524,46 @@ export type SampleHasMetadataSampleMetadataNodeAggregationWhereInput = {
   baitSet_SHORTEST_GTE?: InputMaybe<Scalars["Int"]>;
   baitSet_SHORTEST_LT?: InputMaybe<Scalars["Int"]>;
   baitSet_SHORTEST_LTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_AVERAGE_EQUAL?: InputMaybe<Scalars["Float"]>;
+  cancerTypeDetailed_AVERAGE_GT?: InputMaybe<Scalars["Float"]>;
+  cancerTypeDetailed_AVERAGE_GTE?: InputMaybe<Scalars["Float"]>;
+  cancerTypeDetailed_AVERAGE_LT?: InputMaybe<Scalars["Float"]>;
+  cancerTypeDetailed_AVERAGE_LTE?: InputMaybe<Scalars["Float"]>;
+  cancerTypeDetailed_EQUAL?: InputMaybe<Scalars["String"]>;
+  cancerTypeDetailed_GT?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_GTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LONGEST_EQUAL?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LONGEST_GT?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LONGEST_GTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LONGEST_LT?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LONGEST_LTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LT?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_LTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_SHORTEST_EQUAL?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_SHORTEST_GT?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_SHORTEST_GTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_SHORTEST_LT?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_SHORTEST_LTE?: InputMaybe<Scalars["Int"]>;
+  cancerType_AVERAGE_EQUAL?: InputMaybe<Scalars["Float"]>;
+  cancerType_AVERAGE_GT?: InputMaybe<Scalars["Float"]>;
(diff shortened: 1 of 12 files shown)


graphql-server/src/utils/oncotree.ts
```diff
@@ -0,0 +1,112 @@
...
+  name: string;
+  parent: string;
+  precursors: string[];
+  revocations: string[];
+  tissue: string;
+};
+
+export type CachedOncotreeData =
+  | Pick<OncotreeTumorType, "name" | "mainType">
+  | undefined;
```
qu8n, line 26: TypeScript's `Pick<>` lets me "pick" `name` and `mainType` props from `OncotreeTumorType`. This is equivalent to the following:

```ts
export type CachedOncotreeData = {
    name: string;
    mainType: string;
} | undefined;
```

I wanted to avoid overloading the cache with unnecessary data, keeping it light.

graphql-server/src/utils/servers.ts
```diff
@@ -25,17 +27,27 @@ export function initializeHttpsServer(app: Express) {
...
 ) {
   const neo4jDbSchema = await buildNeo4jDbSchema();
-
   const mergedSchema = mergeSchemas({
     schemas: [neo4jDbSchema, oracleDbSchema],
   });
 
-  const apolloServer = new ApolloServer({
+  const oncotreeCache = new NodeCache({ stdTTL: 86400 }); // 1 day
+  await fetchAndCacheOncotreeData(oncotreeCache);
```
qu8n, line 48: This is the Initial fetch and caching of Oncotree data when the GraphQL server boots up.

frontend/src/generated/graphql.ts
```diff
@@ -7524,6 +7524,46 @@ export type SampleHasMetadataSampleMetadataNodeAggregationWhereInput = {
   baitSet_SHORTEST_GTE?: InputMaybe<Scalars["Int"]>;
   baitSet_SHORTEST_LT?: InputMaybe<Scalars["Int"]>;
   baitSet_SHORTEST_LTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_AVERAGE_EQUAL?: InputMaybe<Scalars["Float"]>;
```
ao508, line 7527: This change is a little surprising to me since we aren't adding these properties to the actual graph database. Are these here because they're part of the extended data types used by the custom resolvers?

graphql-server/src/generated/graphql.ts
```diff
@@ -7523,6 +7523,46 @@ export type SampleHasMetadataSampleMetadataNodeAggregationWhereInput = {
   baitSet_SHORTEST_GTE?: InputMaybe<Scalars["Int"]>;
   baitSet_SHORTEST_LT?: InputMaybe<Scalars["Int"]>;
   baitSet_SHORTEST_LTE?: InputMaybe<Scalars["Int"]>;
+  cancerTypeDetailed_AVERAGE_EQUAL?: InputMaybe<Scalars["Float"]>;
```
ao508, line 7526: same comment as [here](https://github.com/mskcc/smile-dashboard/pull/145/files#r1697504503)

frontend/src/generated/graphql.ts
qu8n: Good call-out. And yes. These changes don't do anything (we can't query these fields against Neo4j), but they get generated as long as these fields are part of our schema. If we want our generated GraphQL types to be extra clean we could run some scripts to remove them every time we re-run the codegen, but I don't think we can remove them from the type generation process entirely.

qu8n: Alternatively, we could potentially avoid this issue by using the [GraphQL CodeGen plugin](https://the-guild.dev/graphql/codegen/docs/custom-codegen/extend-schema) for extending the schema. I'll spend some time with this and let you know if it would be useful.

graphql-server/src/utils/constants.ts
```diff
@@ -36,6 +36,8 @@ export const props = {
 
   web_key_pem: properties.get("web.web_key_pem"),
   web_cert_pem: properties.get("web.web_cert_pem"),
+
+  oncotree_api: properties.get("oncotree.oncotree_api"),
```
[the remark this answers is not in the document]
ao508: Can you add this property to the example props file as well?

frontend/src/generated/graphql.ts
qu8n: I spent some time with extending the schema via the plugin, and it seems a bit buggy right now. I can make a separate ticket for this to spend more time on it if you think that's needed

ao508: Something documenting the addition of these properties due to the custom resolvers should be sufficient. Let's plan to add/update docs on the dashboard models used.

ao508 (approved):

PR #148, opened 2024-08-09
Enable searching `cancerType` and `cancerTypeDetailed` fields in the Samples views

qu8n:
This PR contains the ongoing work for card [#1244](https://app.zenhub.com/workspaces/metadb-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1244).

Progress tracker:
- [x] General Samples page
- [x] Request Samples view
- [x] Patient Samples view
- [x] Cohort Samples view - Nothing to implement here but we should test to make sure nothing breaks

### DIFF
--- a/frontend/src/components/SamplesList.tsx
+++ b/frontend/src/components/SamplesList.tsx
@@ -157,6 +157,7 @@ export default function SamplesList({
       setChanges(updatedChanges);
       if (updatedChanges.length === 0) setUnsavedChanges?.(false);
       resetAlertIfCostCentersAreAllValid(updatedChanges);
+      gridRef.current?.api?.refreshCells({ rowNodes: [rowNode] });
       return;
     }
 
@@ -229,6 +230,7 @@ export default function SamplesList({
     }
 
     setUnsavedChanges?.(true);
+    gridRef.current?.api?.refreshCells({ rowNodes: [rowNode] });
   }
 
   const handleDiscardChanges = () => {
--- a/frontend/src/shared/components/InfoToolTip.tsx
+++ b/frontend/src/shared/components/InfoToolTip.tsx
@@ -1,10 +1,19 @@
 import { Tooltip } from "@material-ui/core";
 import InfoIcon from "@material-ui/icons/InfoOutlined";
+import { withStyles } from "@material-ui/core/styles";
+
+const CustomTooltip = withStyles({
+  tooltip: {
+    backgroundColor: "white",
+    color: "black",
+    border: "1px solid lightgrey",
+  },
+})(Tooltip);
 
 export function InfoToolTip({ children }: { children: React.ReactNode }) {
   return (
-    <Tooltip title={<span style={{ fontSize: 12 }}>{children}</span>}>
+    <CustomTooltip title={<span style={{ fontSize: 12 }}>{children}</span>}>
       <InfoIcon style={{ fontSize: 18, color: "grey" }} />
-    </Tooltip>
+    </CustomTooltip>
   );
 }
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -1044,7 +1044,7 @@ export function getSamplePopupParamId(
   samples: SamplesListQuery["samples"],
   paramId: string
 ) {
-  if (parentWhereVariables.OR?.[0].patientsHasSampleConnection_SOME) {
+  if (parentWhereVariables.OR?.[0].patientsHasSample_SOME) {
     const { cmoPatientId, dmpPatientId } = samples[0] || {};
     if (cmoPatientId) return cmoPatientId;
     if (dmpPatientId) return dmpPatientId;
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -1045,9 +1045,9 @@ export function getSamplePopupParamId(
   paramId: string
 ) {
   if (parentWhereVariables.OR?.[0].patientsHasSampleConnection_SOME) {
-    const { cmoPatientId, dmpPatientId } = samples[0];
+    const { cmoPatientId, dmpPatientId } = samples[0] || {};
     if (cmoPatientId) return cmoPatientId;
-    return dmpPatientId;
+    if (dmpPatientId) return dmpPatientId;
   }
   return paramId;
 }
--- a/graphql-server/src/schemas/neo4j.ts
+++ b/graphql-server/src/schemas/neo4j.ts
@@ -35,6 +35,7 @@ import {
   CachedOncotreeData,
   includeCancerTypeFieldsInSearch,
 } from "../utils/oncotree";
+import { querySamplesList } from "../utils/ogm";
 
 type SortOptions = { [key: string]: SortDirection }[];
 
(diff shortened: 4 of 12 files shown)


frontend/src/shared/components/InfoToolTip.tsx
qu8n, line 1: Previously we used double quotes for the bulk search example ("value1 value2 value3") in the search tooltip. It might be unclear to have doubles quotes nested inside double quotes to demonstrate the new quoted string example, so I switched to wrapping the example with the `inline code style`. This style isn't readable on the default dark gray background, so I added a white background to our tooltip.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -336,109 +342,30 @@ function buildResolvers(
...
+        _source: undefined,
+        args: any,
+        { oncotreeCache }: ApolloServerContext
+      ) {
+        let customWhere = includeCancerTypeFieldsInSearch(
+          args.where,
+          oncotreeCache
+        );
+        return await querySamplesList(ogm, customWhere, args.options);
+      },
```
qu8n, line 355: Before querying the database, we will modify the `where`/filter to include oncotree codes corresponding to the `ct`/`ctDetailed` values from users' search input.

Moved the database call into the named function `querySamplesList` because we have to use it again for the `samplesConnection` resolver below.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -336,109 +342,30 @@ function buildResolvers(
...
+        { oncotreeCache }: ApolloServerContext
+      ) {
+        let customWhere = includeCancerTypeFieldsInSearch(
+          args.where,
+          oncotreeCache
+        );
+        const samples = await querySamplesList(ogm, customWhere, args.options);
+        return {
+          totalCount: samples.length,
+        };
```
qu8n, line 368: Note that because we're calling `includeCancerTypeFieldsInSearch` 2x for each samples query, it's going to be 2x slower than before. We can discuss mitigating with the upcoming samples cache implementation.

graphql-server/src/utils/oncotree.ts
qu8n, line 1: This function includes cancer type fields in the search query indirectly by searching the Oncotree cache for matching cancerType or cancerTypeDetailed values. If a match is found, it adds the corresponding Oncotree code to the original search query, which is passed to the database query.

graphql-server/src/utils/oncotree.ts
```diff
@@ -110,3 +112,50 @@ async function getOncotreeCodesFromNeo4j() {
...
+export function includeCancerTypeFieldsInSearch(
+  where: SampleWhere,
+  oncotreeCache: NodeCache
+) {
+  const customWhere = { ...where };
+
+  const sampleMetadataFilters =
+    customWhere?.hasMetadataSampleMetadata_SOME?.OR ||
+    customWhere?.OR?.find((filter) => filter.hasMetadataSampleMetadata_SOME)
+      ?.hasMetadataSampleMetadata_SOME?.OR;
```
qu8n, line 130: Extracts the filters. The resulting `sampleMetadataFilters` might look like this, for example:
```js
[
  {
    "cmoSampleName_CONTAINS": "C-8W908N"
  },
  {
    "importDate_CONTAINS": "C-8W908N"
  },
  ...
]
```

Because each view's `where` is structured differently, we handle each case separately in a logical OR chain. Given `const var = x || y || z;`; if `x` is truthy, `var` gets assigned `x`; if `x` is falsy, we check the truthiness of `y`, and so on.

In this code, `customWhere?.hasMetadataSampleMetadata_SOME?.OR` handles the Request Samples view, and the latter handles the general Samples page. We'll need modify this line to handle the Patient Samples view and potentially the Cohort Samples view.

graphql-server/src/utils/oncotree.ts
```diff
@@ -110,3 +112,50 @@ async function getOncotreeCodesFromNeo4j() {
...
+
+  if (sampleMetadataFilters?.length) {
+    const searchInput = Object.values(sampleMetadataFilters[0])[0] as
+      | string
+      | string[];
+    const searchValues = Array.isArray(searchInput)
+      ? searchInput
+      : [searchInput];
```
qu8n, line 138: Extract the user input. `searchValues` will equal `["someValue"]` for a single-value search, and `["val1", "val2", "val3", ...]` for a bulk search.

graphql-server/src/utils/oncotree.ts
```diff
@@ -110,3 +112,50 @@ async function getOncotreeCodesFromNeo4j() {
...
+      if (
+        searchValues.some(
+          (val) =>
+            name?.toLowerCase().includes(val.toLowerCase()) ||
+            mainType?.toLowerCase().includes(val.toLowerCase())
+        )
+      ) {
+        addOncotreeCode(code);
+      }
+    });
```
qu8n, line 157: Iterate through each Oncotree code in the cache. For each code, look at its respective `name` and `mainType` values. If either partially matches the user input, we'll add the corresponding Oncotree code to the `where`/filter that will be included in the database query.

frontend/src/shared/components/InfoToolTip.tsx
qu8n, line 1: Previously we used double quotes for the bulk search example ("value1 value2 value3") in the search tooltip. It might be unclear to have doubles quotes nested inside double quotes to demonstrate the new quoted string example, so I switched to wrapping the example with the `inline code style`. This style isn't readable on the default dark gray background, so I added a white background to our tooltip.

graphql-server/src/utils/oncotree.ts
```diff
@@ -110,3 +112,50 @@ async function getOncotreeCodesFromNeo4j() {
     await session.close();
   }
 }
+
+/**
+ * Include cancer type fields in the search query indirectly by searching the Oncotree cache
+ * for matching cancerType or cancerTypeDetailed values. If a match is found, add the corresponding
+ * Oncotree code to the original search query.
+ */
+export function includeCancerTypeFieldsInSearch(
```
qu8n, line 121: This function includes cancer type fields in the search query indirectly by searching the Oncotree cache for matching cancerType or cancerTypeDetailed values. If a match is found, it adds the corresponding Oncotree code to the original search query, which is passed to the database query.

frontend/src/components/SamplesList.tsx
```diff
@@ -229,6 +230,7 @@ export default function SamplesList({
     }
 
     setUnsavedChanges?.(true);
+    gridRef.current?.api?.refreshCells({ rowNodes: [rowNode] });
```
qu8n, line 233: I noticed that cells are not "yellow" after being updated the first time, only the second time.

The problem is that this `unsubmittedChange` styling class isn't applied immediately after a cell value changes because the state update is asynchronous, and AG Grid might re-render the cell before the state is updated.

These lines force AG Grid to re-render the row right after updating the state, ensuring the class rule is applied correctly.

frontend/src/shared/helpers.tsx
```diff
@@ -1044,10 +1044,10 @@ export function getSamplePopupParamId(
   samples: SamplesListQuery["samples"],
   paramId: string
 ) {
-  if (parentWhereVariables.OR?.[0].patientsHasSampleConnection_SOME) {
-    const { cmoPatientId, dmpPatientId } = samples[0];
+  if (parentWhereVariables.OR?.[0].patientsHasSample_SOME) {
+    const { cmoPatientId, dmpPatientId } = samples[0] || {};
     if (cmoPatientId) return cmoPatientId;
-    return dmpPatientId;
+    if (dmpPatientId) return dmpPatientId;
```
qu8n, line 1050: Safer handling of cases where `samples[0]` is undefined or `dmpPatientId` doesn't exist.

ao508 (approved):

PR #149, opened 2024-08-23
Filter samples view for only WES samples when clicking the 'View WES samples' tab

qu8n:


frontend/src/components/SamplesList.tsx
```diff
@@ -109,8 +109,7 @@ export default function SamplesList({
     refetchSearchVal().then(() => {
       gridRef.current?.api?.hideOverlay();
     });
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [parsedSearchVals]);
+  }, [parsedSearchVals, columnDefs, refetchWhereVariables, refetch]);
```
qu8n, line 112: Updated the variable dependencies that trigger the refetching of data. The most important one here being `columnDefs` which defines the columns shown to users. So whenever users click on the "View WES samples" tab, the dashboard will refetch (currently, it does not do that).

frontend/src/pages/samples/SamplesPage.tsx
```diff
@@ -28,14 +38,19 @@ export default function SamplesPage() {
         };
         return {
           OR: cohortSampleFilters.concat(sampleMetadataFilters),
-        } as SampleWhere;
+          ...(_.isEqual(columnDefs, ReadOnlyCohortSampleDetailsColumns) && {
+            hasMetadataSampleMetadata_SOME: {
+              OR: sampleFilterWhereVariables(WES_SAMPLE_FILTERS),
+            },
+          }),
+        };
```
qu8n, line 46: When users click on a tab, if that tab is "View WES samples", we will include this extra "filter" with the WES criteria. This filter falls outside the `OR` clause so it will always be included in subsequent queries/searches.

ao508 (approved):

PR #150, opened 2024-08-26
Bug fix - data sanitization on export

ao508:


frontend/src/utils/stringBuilders.ts
```diff
@@ -25,7 +25,7 @@ export function buildTsvString(rows: any[], colDefs: ColDef[]) {
         return " ";
       })
     )
-    .map((value) => value.join("\t"));
+    .map((value) => value.join("\t").replace(/(\r\n|\n|\r)/gm, ""));
```
ao508, line 28: data cleanup happening on a per-line basis instead of each datum of the row

qu8n (approved):

PR #152, opened 2024-08-30
Automatically sort samples query results by importDate

qu8n:
For card [#1225](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1225).

This PR implements the following changes:
- Automatically sort samples query results by importDate in descending order
- Improve performance of the React frontend by removing an unnecessary re-render on the Samples page
- Improve performance of the GraphQL server by batching child queries of the SamplesList query, removing an extra trip to the database

The performance improvements reduced the initial loading time by 2x, but it still takes a while; I'm seeing 1m30s - 2m.

In parallel, we'll add a new environment variable `NODE_OPTIONS="--max-old-space-size=16384"`. This increases the heap memory space available to the Node.js runtime from 4GB to 16GB. Fetching all samples and sorting them in each user request is expensive. This increase will also resolve the error we saw in card [#1264](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1264).

CleanShot 061210

### DIFF
--- a/frontend/src/components/SamplesList.tsx
+++ b/frontend/src/components/SamplesList.tsx
@@ -72,9 +72,6 @@ export default function SamplesList({
     useSamplesListQuery({
       variables: {
         where: parentWhereVariables || {},
-        options: {
-          limit: MAX_ROWS,
-        },
         sampleMetadataOptions: {
           sort: [{ importDate: SortDirection.Desc }],
           limit: 1,
--- a/frontend/src/pages/samples/SamplesPage.tsx
+++ b/frontend/src/pages/samples/SamplesPage.tsx
@@ -24,27 +24,29 @@ const WES_SAMPLE_FILTERS = [
 export default function SamplesPage() {
   const [columnDefs, setColumnDefs] = useState(combinedSampleDetailsColumns);
 
+  const refetchWhereVariables = (parsedSearchVals: string[]) => {
+    const cohortSampleFilters = cohortSampleFilterWhereVariables(
+      parsedSearchVals
+    ).filter((filter) => filter.hasTempoTempos_SOME);
+    const sampleMetadataFilters = {
+      hasMetadataSampleMetadata_SOME: {
+        OR: sampleFilterWhereVariables(parsedSearchVals),
+      },
+    };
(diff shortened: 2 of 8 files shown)


frontend/src/components/SamplesList.tsx
```diff
@@ -72,9 +72,6 @@ export default function SamplesList({
     useSamplesListQuery({
       variables: {
         where: parentWhereVariables || {},
-        options: {
-          limit: MAX_ROWS,
-        },
```
qu8n, line 77: Remove this limit (500 rows) as we need to fetch all samples and then sort them in the GraphQL server. Note that ultimately we'll still return 500 samples to the user.

ao508 (commented):
A couple questions I had while testing this out in dev.. 

After waiting for the samples page to load, I clicked the "View WES samples" button and immediately got an error (Apollo fetch error). However, the error seemed to resolve on its own without doing anything. I was still on the same page, didn't refresh or anything, and it seemed the data just refreshed itself. I'm just curious if you have seen this happen as well or maybe it's just a hiccup on my side.

I'm wondering if it also makes sense to include the import date on the WES view? Let's ask the PMs if this is helpful to them. I only thought about it now that the data is getting sorted by import date.

qu8n:
> A couple questions I had while testing this out in dev..
> 
> After waiting for the samples page to load, I clicked the "View WES samples" button and immediately got an error (Apollo fetch error). However, the error seemed to resolve on its own without doing anything. I was still on the same page, didn't refresh or anything, and it seemed the data just refreshed itself. I'm just curious if you have seen this happen as well or maybe it's just a hiccup on my side.
> 
> I'm wondering if it also makes sense to include the import date on the WES view? Let's ask the PMs if this is helpful to them. I only thought about it now that the data is getting sorted by import date.

I was able to replicate the fetch error and am seeing an unfamiliar error message in the GraphQL server's logs. Will keep you posted on this.

ao508 (approved):

PR #153, opened 2024-09-05
Increase the sample export limit

qu8n:
For [card 1237](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1237). This PR increases the max number of samples that can be exported from 500 to 5,000.

If users attempt to export more, they will see a popup asking them to either refine their search or reach out to our team.

### DIFF
--- a/frontend/src/components/SamplesList.tsx
+++ b/frontend/src/components/SamplesList.tsx
@@ -40,7 +40,7 @@ const MAX_ROWS_EXPORT = 5000;
 const MAX_ROWS_SCROLLED_ALERT =
   "You've reached the maximum number of samples that can be displayed. Please refine your search to see more samples.";
 const MAX_ROWS_EXPORT_EXCEED_ALERT =
-  "You can only download up to 5,000 rows of data at a time. Please refine your search and try again. If you need the full dataset, contact the SMILE team.";
+  "You can only download up to 5,000 rows of data at a time. Please refine your search and try again. If you need the full dataset, contact the SMILE team at cmosmile@mskcc.org";
 const COST_CENTER_VALIDATION_ALERT =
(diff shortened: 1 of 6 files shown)


frontend/src/components/SamplesList.tsx
```diff
@@ -90,26 +96,23 @@ export default function SamplesList({
...
   const params = useParams();
 
   useEffect(() => {
     gridRef.current?.api?.showLoadingOverlay();
-    async function refetchSearchVal() {
-      await refetch({
-        where: refetchWhereVariables(parsedSearchVals),
-      });
-    }
-    refetchSearchVal().then(() => {
```
qu8n, line 106: Feels good to actually reduce a bit of code here. `refetch` is not an async function so we didn't need to `async/await` it. Rather, it's a regular function that returns a promise, which is slightly different.

frontend/src/components/SamplesList.tsx
```diff
@@ -262,10 +265,25 @@ export default function SamplesList({
...
-            return Promise.resolve(buildTsvString(samples!, columnDefs));
+            return sampleCount <= MAX_ROWS_TABLE
+              ? Promise.resolve(buildTsvString(samples!, columnDefs))
+              : refetch({
+                  options: {
+                    limit: MAX_ROWS_EXPORT,
+                  },
+                }).then((result) =>
+                  buildTsvString(result.data.samples, columnDefs)
+                );
```
qu8n, line 276: If the already-downloaded rows are <= 500 rows, we can just download them directly without refetching.

ao508 (approved):
tested export of < 500 samples and export with > 500 samples but < 5k - both seem to work as intended

PR #155, opened 2024-09-06
Fix displayed consent part a/c status on patient table

ao508:
Fixes issue where records in the Patient view weren't showing 12-245 Part A/C consent statuses even though they exist in the database. 

Current behavior:
image

Expected behavior / behavior after changes from PR:

image

### DIFF
--- a/graphql-server/src/utils/flattening.ts
+++ b/graphql-server/src/utils/flattening.ts
@@ -173,19 +173,25 @@ const nestedValueGetters: NestedValueGetters = {
           ?.join(", ");
       case "consentPartA":
         try {
-          return JSON.parse(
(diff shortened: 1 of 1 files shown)


qu8n (commented):
Instead of `filter`-ing for all truthy values and then grabbing the first one, what do you think about using [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead? Won't make a huge difference here for the Patients page but I'm just trying to be more performance-minded given the current loading time on the Samples page 😅. This looks good otherwise! Thank you

qu8n (approved):
Thanks!

PR #157, opened 2024-09-10
Fix sample status showing warning icon when validationStatus is null

qu8n:
For [card 1268](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1268)

CleanShot 153825

### DIFF
--- a/frontend/src/shared/helpers.tsx
+++ b/frontend/src/shared/helpers.tsx
@@ -224,7 +224,11 @@ export const SampleMetadataDetailsColumns: ColDef[] = [
     headerName: "Status",
     cellRenderer: (params: ICellRendererParams) => {
       if (params.data?.revisable) {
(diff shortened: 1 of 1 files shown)


frontend/src/shared/helpers.tsx
```diff
@@ -224,7 +224,11 @@ export const SampleMetadataDetailsColumns: ColDef[] = [
     headerName: "Status",
     cellRenderer: (params: ICellRendererParams) => {
       if (params.data?.revisable) {
-        return params.data?.validationStatus ? <CheckIcon /> : <WarningIcon />;
```
qu8n, line 227: `validationStatus` can be either `true`, `false`, or `null`. There is a validation report to show only when `validationStatus === false`. This old code incorrectly renders a warning icon when `validationStatus === null`.

ao508 (approved):

PR #161, opened 2024-09-18
Support case-insensitive search for all views

qu8n:
For card [#1263](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1263). This PR enables case-insensitive and partial string search for all views.

Example:
demo


ao508 (dismissed):
Looking good so far! Just aded a reminder that we still want to support the `_CONTAINS` searching

frontend/src/pages/requests/RequestsPage.tsx
```diff
@@ -14,123 +14,30 @@ import {
...
-      { qcAccessEmails_MATCHES: `(?i)${parsedSearchVals[0]}.*` },
-      { dataAccessEmails_MATCHES: `(?i)${parsedSearchVals[0]}.*` },
-      { otherContactEmails_MATCHES: `(?i)${parsedSearchVals[0]}.*` },
-    ];
-  }
-
-  return [];
+  return parsedSearchVals.flatMap((searchVal) =>
+    queryTerms.map((queryTerm) => ({ [queryTerm]: `(?i).*${searchVal}.*` }))
+  );
```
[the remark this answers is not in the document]
qu8n: In this new regex, I added "`.*`" in front of "`${searchVal}`". This new regex will match the substring `searchVal` anywhere in the destination string, essentially doing what `_CONTAINS` does, as well as matching case insensitively.

Does this suffice? Or is there another reason that we should keep the `_CONTAINS`?

frontend/src/pages/cohorts/CohortsPage.tsx
qu8n, line 1: Replacing all instances of predicates `_IN` and `_CONTAINS` with `_MATCHES`. This lets us pass in a regex that enables both case-insensitive and partial string searches.

graphql-server/src/schemas/custom.ts
```diff
@@ -636,6 +636,7 @@ const editableSampleMetadataFields = new Set([
   "collectionYear",
   "sampleOrigin",
   "tissueLocation",
+  "sex",
```
qu8n, line 639: This fixes an unrelated bug that prevents users from editing the `sex` field of SampleMetadata.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -24,16 +19,6 @@ export async function buildNeo4jDbSchema() {
...
-  const httpLink = createHttpLink({
-    uri: "https://localhost:4000/graphql",
-    fetch: fetch,
-  });
-
-  const client = new ApolloClient({
-    link: httpLink,
-    cache: new InMemoryCache(),
-  });
-
```
qu8n, line 36: Removed as these are not being used. (These were used previously to query some samples data as part of the old sample mutation workflow.)

graphql-server/src/schemas/neo4j.ts
```diff
@@ -65,14 +50,25 @@ export async function buildNeo4jDbSchema() {
...
   `;
 
-  const ogm = new OGM({ typeDefs: extendedTypeDefs, driver: neo4jDriver });
+  const features = {
+    filters: {
+      String: {
+        MATCHES: true,
+      },
+    },
+  };
```
qu8n, line 59: FYI I removed `ID: true` inside of `String: { ... }`. That config would let us perform RegEx matching on values of type `ID` ([source](https://neo4j.com/docs/graphql/current/queries-aggregations/filtering/)), which is a use case that we don't have.

ao508: Isn't ID something we were interested in accessing?

qu8n: Yes, we are interested in accessing the ID, but setting `ID: true` here does something else. The docs link I shared above says it enables RegEx matching for value of type ID. For example, if we wanted to search for internal ID with containing value "abc". Maybe I missed something there?

PR #162, opened 2024-09-20
Optimized sample dashboard query

mandawilson:


graphql-server/src/schemas/neo4j.ts
```diff
@@ -164,6 +212,205 @@ export async function buildNeo4jDbSchema() {
...
+          qcCompleteResult: mostRecentQcComplete
+            ? mostRecentQcComplete.properties.result
+            : null,
+          qcCompleteReason: mostRecentQcComplete
+            ? mostRecentQcComplete.properties.reason
+            : null,
+          qcCompleteStatus: mostRecentQcComplete
+            ? mostRecentQcComplete.properties.status
+            : null,
+          cohortsHasCohortSample: [],
```
mandawilson, line 391: Remove this if it is not needed.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -31,6 +31,11 @@ import {
   sortArrayByNestedField,
 } from "../utils/flattening";
 import { ApolloServerContext } from "../utils/servers";
+import {
```
mandawilson, line 34: I don't think these are used now - I was just trying stuff.

graphql-server/src/schemas/neo4j.ts
```diff
@@ -247,7 +494,209 @@ function buildResolvers(
...
+
+    // async samples(
+    //   _source: undefined,
+    //   args: any,
+    //   { samplesLoader }: ApolloServerContext
+    // ) {
+
     Query: {
+      // THIS ONE CAN BE CALLED FROM THE BROWSER
+      async sampleDashboardQuery(
```
mandawilson, line 506: I was calling this from the browser with the following, but in the code we use the above method. Don't leave both!
query ExampleQuery { sampleDashboardQuery { primaryId }}

graphql-server/src/utils/ogm.ts
```diff
@@ -1,123 +1,36 @@
...
-import { SortDirection } from "../generated/graphql";
+import { ApolloServerContext } from "./servers";
+//import { SamplesQueryResult } from "./dataloader";
+import { runQuery } from "../schemas/neo4j";
 
 const MAX_ROWS = 500;
 
 export async function querySamplesList(
   ogm: OGM,
   where: GraphQLWhereArg,
```
mandawilson, line 10: I think we have the oncotree data here, we can use it to set cancerType and cancerTypeDetailed.

PR #163, opened 2024-09-26
Enable search for new samples query

qu8n:
This PR includes the work for cards [#1224](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1224) and [#1281](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1281), with most changes for the former.

This enables search for the new samples query written in #162.

High-level details of the new flow:
- User inputs some value(s) into the search bar and press Enter or click "Search"
- The React frontend calls a new GraphQL query called `DashboardSamples`
- In the Apollo server, the resolvers for this query run Cypher queries and return the results

TODOs in the next PR:
- Re-enable editing
- Re-enable users' ability to download >500 and <5,000 samples. Right now, users can either download up to 500 or not at all
- Resolve the duplicate `tableElements.tsx` files
- Consider implementing or using a 3rd party package for a Cypher query builder
- Fix exported file to show only the non-hidden columns


frontend/src/App.tsx
```diff
@@ -36,7 +36,7 @@ export default function App() {
               <PatientsPage userEmail={userEmail} setUserEmail={setUserEmail} />
             }
           >
-            <Route path=":smilePatientId" />
+            <Route path=":patientId" />
```
ao508, line 39: switching away from the smile patient id?

ao508 (commented):
i'm not finished reviewing but I did want to comment really quickly to suggest that we rename the faster query as "fastLoadSamplesQuery" just to avoid some confusion between the original query and the new one

frontend/src/App.tsx
qu8n: Yep. In the Patient Samples view, I figured out a way to have the URL query param == either the `cmoPatientId` or the `dmpPatientId`. For example, see [https://smile-dev.mskcc.org:3006/patients/C-MP76JR](https://smile-dev.mskcc.org:3006/patients/C-MP76JR). Previously this was the `smilePatientId`.

frontend/src/shared/helpers.tsx
```diff
@@ -790,14 +783,18 @@ export const CohortSampleDetailsColumns: ColDef[] = [
     field: "cancerTypeDetailed",
     headerName: "Cancer Type Detailed",
   },
+  {
```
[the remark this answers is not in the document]
qu8n: Yes, let me check with them.

frontend/src/shared/helpers.tsx
```diff
@@ -151,9 +144,9 @@ export const PatientsListColumns: ColDef[] = [
           variant="outline-secondary"
           size="sm"
           onClick={() => {
-            params.context.navigateFunction(
-              `/patients/${params.data.smilePatientId}`
-            );
+            const patientId =
```
ao508, line 147: I'm going to request that we keep the `smilePatientId` as the path value if only to ensure that we're flexible with what kinds of patient aliases we support in the future (i.e., if we start supporting a datasource that has neither a cmo patient id or dmp patient id as the "primary" patient id)

qu8n:
> i'm not finished reviewing but I did want to comment really quickly to suggest that we rename the faster query as "fastLoadSamplesQuery" just to avoid some confusion between the original query and the new one

@ao508 We're no longer using the original query anywhere. I'm deleting it in this PR. Would you still like to rename this new query?

graphql-server/src/schemas/custom.ts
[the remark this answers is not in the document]
ao508: For sure! I will branch off of this PR at some point today to play around with it a bit. Curious about how we can make further improvements to it.

qu8n:
@ao508 This PR's changes are big. I couldn't add comments sequentially in the `Files changed` tab to intuitively explain everything. (Also, the diffs are based on top of Manda's work and not `master`.) Lmk if you'd like a more in-depth in-person walkthrough. I'd be happy to

graphql/operations.graphql
```diff
@@ -46,118 +46,79 @@ query PatientsList($options: PatientOptions, $where: PatientWhere) {
...
-query SamplesList(
-  $where: SampleWhere
-  $options: SampleOptions
-  $sampleMetadataOptions: SampleMetadataOptions
-  $bamCompletesOptions: BamCompleteOptions
-  $mafCompletesOptions: MafCompleteOptions
-  $qcCompletesOptions: QcCompleteOptions
-) {
-  samplesConnection(where: $where) {
+query DashboardSamples($searchVals: [String], $sampleContext: SampleContext) {
```
ao508, line 49: FastLoadSamples ? open to other names as well

ao508 (changes requested):
Some questions and suggestions

frontend/src/shared/helpers.tsx
qu8n: That makes sense. I also think that if we have a new patient alias (let's call it "someNewPatientAliasId"), we could keep this `patientId` structure and use "someNewPatientAliasId" as the URL query param. For example:
```ts
// This PR right now
const patientId = params.data.cmoPatientId || params.data.dmpPatientId;

// In the future, when we have a new patient alias
const patientId = params.data.cmoPatientId || params.data.dmpPatientId || params.data.someNewPatientAliasId;
```
This way the URL can be more user-friendly and we can still be flexible. What do you think?

graphql/operations.graphql
qu8n: We're no longer using the original query anywhere in the dashboard, as of this PR. Would you still like to rename this new query?

frontend/src/shared/helpers.tsx
ao508: ✔️  discussed and confirmed that any valid patient ID can be manually entered in the URL and the corresponding patient sample's table will load. Ex: patient with alias C-AAAAA and P-99999 will both render the same table of samples. 

Okay to keep this change as long as we keep in mind that we may need to support additional patient aliases in the future.

ao508 (commented):
One thing I've noticed while testing is that all columns get downloaded on the samples page even when there are columns hidden. Is this a TODO for a later card?

Related to this: there still seems to be a cap of 500 samples here when downloading from the samples page

qu8n:
> One thing I've noticed while testing is that all columns get downloaded on the samples page even when there are columns hidden. Is this a TODO for a later card?

Thanks for catching that. Adding it to the next PR's todos now.

 
> Related to this: there still seems to be a cap of 500 samples here when downloading from the samples page
> 
> To replicate: I used search term "BRCNOS" which should return 804 samples (dev server)

This is one of my TODOs for the next PR. You can find this list in this PR's description [here](https://github.com/mskcc/smile-dashboard/pull/163#issue-2551474156) for now. I'll add this list to the next Zenhub card after we finish with this PR.

frontend/src/shared/helpers.tsx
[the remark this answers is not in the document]
qu8n: Following the PMs' input, I moved `investigatorSampleId` to the right of `cmoSampleName`. You can see this change in the dev dashboard. Will follow up with a commit for it later today.

frontend/src/components/SamplesList.tsx
```diff
@@ -25,35 +18,36 @@ import "ag-grid-community/styles/ag-grid.css";
...
 import styles from "./records.module.scss";
 import { getUserEmail } from "../utils/getUserEmail";
 import { openLoginPopup } from "../utils/openLoginPopup";
 import { Title } from "../shared/components/Title";
 import { BreadCrumb } from "../shared/components/BreadCrumb";
 import { useParams } from "react-router-dom";
 import { DataName } from "../shared/types";
+import { parseUserSearchVal } from "../utils/parseSearchQueries";
 
 const POLLING_INTERVAL = 2000;
```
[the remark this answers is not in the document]
qu8n: This is in milliseconds, so this is currently at 2 seconds. I just bumped it up to 5 seconds in the dev dashboard. Let me know if you're still seeing it. If you're able to record a clip of this "blinking", I'm happy to take a look as well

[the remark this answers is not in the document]
qu8n: (If this blinking is a big problem, we can consider using websocket to display real-time sample updates instead. More optimal as we don't continuously refetch for no reason.)

ao508 (approved):
Looking good, thank you!

PR #166, opened 2024-10-22
Implement infinite scrolling for the Sample views

qu8n:
This PR transitions the Sample views' AG Grid [row model](https://www.ag-grid.com/javascript-data-grid/row-models/) from Client-Side to Server-Side, allowing the grid to lazy-load the data as the user scrolls down.

[#1291](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1291)

### DIFF
--- a/frontend/src/components/SamplesList.tsx
+++ b/frontend/src/components/SamplesList.tsx
@@ -121,8 +121,10 @@ export default function SamplesList({
           .then((result) => {
             successCallback(
               result.data.dashboardSamples,
-              result.data.dashboardSampleCount.totalCount ?? 0
+              result.data.dashboardSampleCount.totalCount
             );
+
+            setSampleCount(result.data.dashboardSampleCount.totalCount);
           })
           .catch(() => {
             failCallback();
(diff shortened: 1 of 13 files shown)


frontend/src/components/SamplesList.tsx
```diff
@@ -186,33 +235,59 @@ export default function SamplesList({
       return [...changes];
     });
 
-    // validate Cost Center inputs
-    if (fieldName === "costCenter") {
-      if (!isValidCostCenter(newValue)) {
-        setAlertContent(COST_CENTER_VALIDATION_ALERT);
-        setShowAlertModal(true);
-      } else {
-        resetAlertIfCostCentersAreAllValid(changes);
```
ao508, line 195: Do we still need this behavior or not with the ways things have been refactored?

ao508 (commented):
A lot of complex logic and changes in SamplesList.tsx but I think I've wrapped my head around it for the most part. Let's walk through it tomorrow during our 1:1

frontend/src/components/SamplesList.tsx
qu8n: Good question -- We do. This behavior is still happening. I simplified the React.js states for displaying alert modals, so the new code is shorter but is producing the same behavior.

ao508 (approved):

PR #167, opened 2024-11-01
Migrate Patients page to using a direct Neo4j query

qu8n:
For card [#1292](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1292). This PR migrates the Patients page from OGM to using a direct Neo4j query, improving the performance of the page.

### DIFF
--- a/graphql-server/src/schemas/custom.ts
+++ b/graphql-server/src/schemas/custom.ts
@@ -78,7 +78,7 @@ export async function buildCustomSchema(ogm: OGM) {
       ) {
         const queryBody = buildPatientsQueryBody({ searchVals, filter });
         return await queryDashboardPatients({
-          queryBody: queryBody,
+          queryBody,
           sort,
           limit,
           offset,
@@ -89,9 +89,7 @@ export async function buildCustomSchema(ogm: OGM) {
         { searchVals, filter }: QueryDashboardPatientCountArgs
       ) {
         const queryBody = buildPatientsQueryBody({ searchVals, filter });
-        return await queryDashboardPatientCount({
-          queryBody,
-        });
+        return await queryDashboardPatientCount({ queryBody });
       },
     },
     Mutation: {
--- a/graphql-server/src/schemas/custom.ts
+++ b/graphql-server/src/schemas/custom.ts
@@ -368,7 +368,7 @@ async function queryDashboardSamples({
       latestQC.reason AS qcCompleteReason,
       latestQC.status AS qcCompleteStatus
 
-    ORDER BY ${sort.colId} ${sort.sort}
+    ORDER BY ${getNeo4jCustomSort(sort)}
     SKIP ${offset}
     LIMIT ${limit}
   `;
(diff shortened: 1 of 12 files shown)


graphql-server/src/schemas/custom.ts
```diff
@@ -323,7 +366,7 @@ async function queryDashboardSamples({
       latestQC.reason AS qcCompleteReason,
       latestQC.status AS qcCompleteStatus
 
-    ORDER BY ${sort.colId} ${sort.sort}
+    ORDER BY ${getNeo4jCustomSort(sort)}
```
ao508, line 369: 👍

graphql-server/src/schemas/custom.ts
```diff
@@ -758,3 +801,150 @@ async function publishNatsMessage(topic: string, message: string) {
...
+
+    // Implode/aggregate the different arrays
+    WITH
+      p,
+      cmoPa,
+      dmpPa,
+      collect(s) as samples,
+      collect(cmoSampleId) AS cmoSampleIds,
+      collect(latestSmAddlPropsJson.\`consent-parta\`) as consentPartAs,
+      collect(latestSmAddlPropsJson.\`consent-partc\`) as consentPartCs
```
ao508, line 871: I assume this will return null if these fields don't exist in the additional properties? Can you confirm?

graphql-server/src/schemas/custom.ts
```diff
@@ -758,3 +801,150 @@ async function publishNatsMessage(topic: string, message: string) {
...
+      dmpPa,
+      s,
+      collect(sm) AS allSampleMetadata,
+      max(sm.importDate) AS latestImportDate
+    WITH
+      p,
+      cmoPa,
+      dmpPa,
+      s,
+      [sm IN allSampleMetadata WHERE sm.importDate = latestImportDate][0] AS latestSm
```
ao508, line 849: Looks like we are collecting info on what the latest sample metadata is so maybe it _is_ possible to accomplish the "sort by date" after all? As discussed previously though this is not something that's hard required. It would just be a nice bonus. Worth a shot

graphql-server/src/schemas/custom.ts
qu8n: Good question. Confirmed. Accessing non-existent/made-up properties in a node returns `null` ([source](https://neo4j.com/developer/kb/understanding-non-existent-properties-and-null-values/)).

If the only input to `collect( )` is `null`, as in `collect(null)`, then it returns an empty array `[ ]`.

Going some lines below this section, we have `consentPartAs[0]`. If `consentPartAs` is an empty array `[ ]`, then accessing an index on it also returns a `null`.

graphql-server/src/schemas/custom.ts
qu8n: Yep, this is possible. I'll Slack you the results.

ao508 (approved):

PR #175, opened 2025-01-27
Add missing sample validation info in the samples query

qu8n:
This PR adds back the missing validation info that is missing from the new samples query.

CleanShot 165600

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,8 +40,13 @@ workflows:
   version: 2
   install_and_prettify:
     jobs:
-      - install
+      - install:
+          filters:
+            branches:
+              ignore: master
       - prettier:
           requires:
             - install
-
+          filters:
+            branches:
+              ignore: master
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -40,8 +40,13 @@ workflows:
...
+          filters:
+            branches:
+              ignore: master
       - prettier:
           requires:
             - install
-
+          filters:
+            branches:
+              ignore: master
```
qu8n, line 52: This fixes the CircleCI error whenever we merge PRs to master:
CleanShot 160936

I propose that we don't run this script (`prettierCheckCircleCI`) when merging to master because it's already a part of a pull request's checks. Running it again might be redundant. What do you think?

Here's what I think causes this error:
- When running on master, the `prettierCheckCircleCI` script is diffing master against itself, which results in no changes
- When there are no changes, the last part of the script won't be executed (`git remote remove smile-repo`)
- So when this script runs after the first time, it fails in `master` because `smile-repo` remote already exists

We can confirm the above by running `prettierCheckCircleCI` in the local master branch.

qu8n:
These changes have been deployed to dev.

ao508 (approved):

PR #178, opened 2025-03-14
Handle empty search results

ao508:


frontend/src/components/RecordsList.tsx
```diff
@@ -91,8 +91,14 @@ export default function RecordsList({
 
   const dataNameCapitalized = buildSentenceCaseString(dataName);
   const recordsQueryName = `dashboard${dataNameCapitalized}`;
-  const recordCount = data?.[recordsQueryName][0]._total;
-  const uniqueSampleCount = data?.[recordsQueryName][0]._uniqueSampleCount;
```
qu8n, line 95: Nice. JS has built-in operators specifically for this scenario. Could we use those instead? For example:
```ts
  const recordCount = data?.[recordsQueryName][0]?._total || 0;
  const uniqueSampleCount = data?.[recordsQueryName][0]?._uniqueSampleCount || 0;
```

frontend/src/components/RecordsList.tsx
```diff
@@ -133,7 +139,10 @@ export default function RecordsList({
             .then((result) => {
               params.success({
                 rowData: result.data[recordsQueryName],
-                rowCount: result.data?.[recordsQueryName][0]?._total,
```
qu8n, line 136: Similar to the other comment. (And the same for `SamplesList.tsx`)
```ts
                rowCount: result.data?.[recordsQueryName][0]?._total || 0,
```

qu8n (changes requested):
Thanks! I left a small change request to leverage JS' neat built-in operators.

frontend/src/components/SamplesList.tsx
```diff
@@ -96,10 +96,7 @@ export default function SamplesList({
     });
 
   const samples = data?.dashboardSamples;
-  const sampleCount =
-    data?.dashboardSamples[0] !== undefined
-      ? data?.dashboardSamples[0]._total
-      : 0;
+  const sampleCount = data?.dashboardSamples[0]._total || 0;
```
qu8n, line 99: Could we add the missing `?.` operator here? We can merge otherwise. Thanks!
```ts
const sampleCount = data?.dashboardSamples[0]?._total || 0;
```

qu8n (approved):

frontend/src/components/SamplesList.tsx
qu8n: So nothing breaks when there is 0 records returned.

ao508: added

qu8n (approved):

PR #180, opened 2025-03-25
Cache initial results of the Samples page

qu8n:
This PR made the following changes:
- Cache the results of the initial queries on the Samples page (Zenhub ticket [#1453](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1453))
- Fix a bug that caused duplicate fetches when refreshing cache, leading to cache errors (Zenhub ticket [#1449](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1449))
- Add timestamps to console logs to ease future debugging

This PR reduces the % memory usage from 44.5 to 43.8 (seen in dev).


graphql-server/src/utils/cache.ts
```diff
@@ -0,0 +1,222 @@
...
+    "CustomCapture",
+    "IDT_Exome_v1_FP",
+    "IDT_Exome_V1_IMPACT468",
+    "WES_Human",
+    "WholeExomeSequencing",
+  ],
+};
+
+export type OncotreeCache = Record<string, { name: string; mainType: string }>; // key = Oncotree code
+export type SamplesCache = Record<string, DashboardSample[]>; // key = full Cypher query
```
qu8n, line 40: Current cache ≈ an object/hashmap where each key is an oncotree code and corresponding value contains the cancer type info

New cache ≈ 
```ts
{
    "oncotree": OncotreeCache,
    "samples": SamplesCache
}
```

graphql-server/src/utils/cache.ts
qu8n, line 1: Many parts in this file, especially around setting up of the Oncotree cache, comes from the old `oncotree.ts`. I renamed that file to this `cache.ts` to include all cache-related code.

See the current [oncotree.ts](https://github.com/mskcc/smile-dashboard/blob/master/graphql-server/src/utils/oncotree.ts) for reference.

graphql-server/src/utils/cache.ts
```diff
@@ -0,0 +1,218 @@
...
+  tissue: string;
+};
+
+export async function initializeInMemoryCache() {
+  const inMemoryCache = new NodeCache();
+
+  // Warm up the cache
+  await updateOncotreeCache(inMemoryCache);
+  await updateSamplesCache(inMemoryCache);
+  logCacheStats(inMemoryCache);
```
qu8n, line 66: Logging stats on the cache's memory size upon server startup and cache refreshes so we can monitor its impact on the dev/prod machine's memory.

graphql-server/src/utils/cache.ts
```diff
@@ -0,0 +1,218 @@
...
+      .map((record) => record.get("oncotreeCode"))
+      .filter(Boolean) as string[];
+  } catch (error) {
+    if (error instanceof Error) {
+      console.error(error.message);
+    }
+  } finally {
+    await session.close();
+  }
+}
```
qu8n, line 146: Some simplification of code but most of this remains the same as it was in `oncotree.ts`.

qu8n:
@ao508 This PR has been deployed to the dashboard in dev for your review.

graphql-server/src/utils/cache.ts
```diff
@@ -0,0 +1,218 @@
...
+  values: [
+    "Agilent_51MB",
+    "Agilent_v4_51MB_Human",
+    "CustomCapture",
+    "IDT_Exome_v1_FP",
+    "IDT_Exome_V1_IMPACT468",
+    "WES_Human",
+    "WholeExomeSequencing",
+  ],
+};
```
ao508, line 37: This has me wondering if we can define a global constant in neo4j that `codegen` will automatically generate the react types for that we can use in the frontend/backend dashboard as needed similar to how we're already using the types we define for the `DashboardSample`, etc.

ao508: Not a requirement, just a thought! Trying to think of ways that we can reduce the number of places we'd need to maintain consistency between the two areas of the code.

ao508 (commented):
I'm not sure what the right approach is for addressing discrepancies because it _is_ a little awkward. 

My flow for updating sample `08039_JQ_6`
1. I edited a sample in the samples view (`/samples`) 

As noted already in discussions, this update isn't going to show up on the `/samples` page until the cache refreshes
image


But it _will_ show up if you search for the sample in the `/samples` view

image


And then if you were to immediately clear your search the table will go back to showing the outdated information

image


The options I can think of now are
1. Put a pin in this issue for now and just communicate with the PMs that this discrepancy is going to exist for the time being
2. Find a way to inject the update(s) to the sample(s) in the existing cache so as to make it seem as if the cache really is up to date
3. Trigger a refresh when a sample update happens

graphql-server/src/utils/cache.ts
```diff
@@ -0,0 +1,257 @@
...
+  );
+
+  // Update the fields of any samples in cache that were changed in newDashboardSamples
+  for (const sample of Object.values(samplesCache).flat()) {
+    if (newSamplesByPrimaryId.hasOwnProperty(sample.primaryId)) {
+      const newDashboardSample = newSamplesByPrimaryId[sample.primaryId];
+      for (const field of newDashboardSample.changedFieldNames) {
+        if (field in sample && field in newDashboardSample) {
+          (sample as any)[field] =
+            newDashboardSample[field as keyof DashboardSampleInput];
```
qu8n, line 239: I casted `sample`'s type to `any` because Typescript couldn't easily infer that `field` exists inside `sample`, even though this is inside an if statement that checks that `field` is in `sample`. This isn't the cleanest Typescript approach but it's safe.

qu8n:
> I'm not sure what the right approach is for addressing discrepancies because it _is_ a little awkward.
> 
> My flow for updating sample `08039_JQ_6`
> 
> 1. I edited a sample in the samples view (`/samples`)
> 
> As noted already in discussions, this update isn't going to show up on the `/samples` page until the cache refreshes image
> 
> But it _will_ show up if you search for the sample in the `/samples` view
> 
> image
> 
> And then if you were to immediately clear your search the table will go back to showing the outdated information
> 
> image
> 
> **The options I can think of now are**
> 
> 1. Put a pin in this issue for now and just communicate with the PMs that this discrepancy is going to exist for the time being
> 2. Find a way to inject the update(s) to the sample(s) in the existing cache so as to make it seem as if the cache really is up to date
> 3. Trigger a refresh when a sample update happens
> 4. ?????
> 
> Quan's notes: We discussed and aligned on option 2 as it seems like the most performant option.

@ao508 This has been fixed using option 2, deployed to dev, and tested with multiple sample updates.

ao508 (changes requested):
I think it's very close to fixing that awkward behavior we noticed yesterday but still not quite there. 

I updated the sample type for this sample to Normal and while the sample type field shows that update, the cmo sample name does not (I confirmed that the db entry did get an updated cmo sample name as a result of the update I made)

image

The issue is that we know what update the user wants to make on `Submit` ... but we technically don't know what the label is going to. be updated to (if it's updated). 

Maybe this is a separate mechanism that we need to incorporate? Something that communicates that label update separately?

Actually while we're thinking about that.. what if the status and validation report become "failed"? I think we'd see a similar issue there..

We probably need to brainstorm this a bit more. One thing I can sorta imagine bridging this weird gap in data updates is to schedule a delayed query to neo4j with by the sample primary ids that we are expecting an update for? 

Edit: To add - I think these are the only two pieces of info (`cmoSampleName` and `status`) that we're going to have to handle this way. Maybe we can even build a small query that just returns these two?

graphql-server/src/schemas/queries/samples.ts
```diff
@@ -371,23 +372,118 @@ export async function queryDashboardSamples({
...
+  return Array.from(addlOncotreeCodes);
+}
+
+export type SampleDataForCacheUpdate = Record<
+  string,
+  Pick<
+    DashboardSample,
+    | "primaryId"
+    | "cmoSampleName"
+    | "historicalCmoSampleNames"
```
ao508, line 429: good catch! I forgot about having to include this field in the query for latest updates.

ao508 (approved):
lgtm, thanks!

PR #185, opened 2025-04-18
Extend Samples query data with Instrument Model and Platform

qu8n:
This PR extends Samples query data with Instrument Model and Platform values.

Confirmed that the Samples query returns these new values:
CleanShot 2025-04-17 at 14 12 33

Confirmed that it returns the correct Instrument Model based on the latest run date:
CleanShot 2025-04-17 at 14 09 25

[#1487](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1487)


frontend/src/pages/cohorts/CohortsPage.tsx
```diff
@@ -6,8 +6,8 @@ import { Dispatch, SetStateAction, useState } from "react";
 import {
   MAX_ROWS_EXPORT,
   MAX_ROWS_EXPORT_WARNING,
-  WesSampleDetailsColumns,
-  CohortsListColumns,
+  WesSampleColumns,
+  CohortColumns,
```
qu8n, line 10: Renamed and simplified all column definition variable names.
1. `SampleMetadataDetailsColumns` is incorrect as it includes data beyond SampleMetadata
2. Names like `CohortsListColumns` reflect the old GraphQL query names (`query CohortsList` in this case)

graphql-server/src/schemas/queries/samples.ts
```diff
@@ -386,10 +389,15 @@ export async function queryDashboardSamples({
...
       const recordObject = record.toObject().resultz;
+      const instrumentModel = getInstrumentModelByLibraries(
+        recordObject.libraries
+      );
       return {
         ...recordObject,
         cancerType: oncotreeCache?.[recordObject.oncotreeCode]?.mainType,
         cancerTypeDetailed: oncotreeCache?.[recordObject.oncotreeCode]?.name,
+        instrumentModel: instrumentModel,
+        platform: getPlatformByInstrumentModel(instrumentModel),
```
qu8n, line 400: I put this getter logic here instead of in `frontend/` for better separation of concerns. Ideally we'll keep the data processing logic in the backend resolver and leave the frontend for UI rendering if possible.

graphql-server/src/schemas/queries/samples.ts
```diff
@@ -492,3 +500,73 @@ export async function querySelectSampleDataForCacheUpdate(
...
+      for (const currRun of currLib.runs) {
+        if (!currRun.runMode || !currRun.runDate) continue;
+        const currRunDate = new Date(currRun.runDate);
+        if (!latestRun.runDate || currRunDate > latestRun.runDate) {
+          latestRun.runDate = currRunDate;
+          latestRun.runMode = currRun.runMode;
+        }
+      }
+    }
+    return latestRun.runMode;
```
qu8n, line 534: This block gets the most recent instrument model based on run dates.

In terms of implementation, I thought about combining JS's standard array methods like `flatMap`, `filter`, `reduce` here. Instead, I found this for-loop approach more efficient as we only iterate over each item once. Wanted to avoid adding any more time on this samples query than necessary.

graphql-server/src/utils/typeDefs.ts
qu8n, line 1: This file's content is mostly cut and pasted from `custom.ts`, with two changes:
1. Typedefs are organized into different constants for ease of reading.
2. We no longer have to define props in `DashboardSample` and `DashboardSampleInput` types twice.

ao508 (approved):

PR #195, opened 2025-05-13
Enable viewing request-level validation reports

qu8n:
This PR implements the following changes:
- Add the ability to view Request-level validation reports in a popup modal
- Update the validation viewing UX in Samples views to be consistent with Requests page's

### Screenshots

Hovering over the warning icon shows a tooltip:
CleanShot 2025-05-13 at 16 38 19

Request validation modal example:
CleanShot 2025-05-13 at 17 36 47

[#1079](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1079)

Signed-off-by: Quan Nguyen <86090707+qu8n@users.noreply.github.com>

PR #198, opened 2025-05-15
Enable viewing request-level validation reports (#195)

ao508:
* Simplify requests query's list of fields to search

* Update requests query to fetch only the latest RequestMetadata

* Update the requests query to return validationReport and validationStatus

* Add a Status column to the Requests page

* Show tooltip upon request status warning icon hover

* Open a modal when clicking on the request warning icon

* Fill in request validation data onto the new modal

* Display sample validation errors in the request validation modal

* Update the validation-action details

* Group sample-level validations under a header

* Make sample validation report a modal to be consistent

* Remove red background for samples with validation status to be consistent

* Re-enable samples caching on server startup

* Fix request validation not showing all sample validation errors

---------


qu8n (approved):
Looks great. Thank you for wrapping up this work!

PR #221, opened 2025-07-09
Only show unique historical CMO sample names

ao508:


graphql-server/src/schemas/queries/samples.ts
```diff
@@ -182,9 +182,12 @@ export function buildSamplesQueryBody({
...
       COLLECT {
       	MATCH (s)-[:HAS_METADATA]->(sm:SampleMetadata)
+        WITH 
+          sm.cmoSampleName as cmoSampleName,
+          apoc.coll.sort(COLLECT(sm.importDate)) as orderedImportDates
       	RETURN ({
-      		cmoSampleName: sm.cmoSampleName,
-      		importDate: sm.importDate
+      		cmoSampleName: cmoSampleName,
+      		importDate: orderedImportDates[-1]
```
ao508, line 190: `apoc.coll.sort` only takes one argument (a list) and doesn't, to my knowledge at least, support reversing the collection. We might need to upgrade apoc to get the version with `apoc.coll.reverse` but I think that's for another time.

qu8n (approved):

PR #238, opened 2025-08-06
Add anchor OncoTree code to patients page

ao508:
Changes:
- anchor oncotree code was added to the patients view schema
- decided against adding anchor oncotree code as a searchable term due to the significant performance hit on the patients view page

TO-DO:
- [x] confirm download works (with and without phi)
- [x] confirm search works on main patient view page

image


qu8n (approved):
The code changes all make sense. I left a nit comment and am ok with the current state. Let me know if you'd like my help with reviewing the changes in dev.

qu8n (approved):

PR #241, opened 2025-08-22
Add tooltips to select whole exome sample columns

qu8n:
## Description

- Add tooltips to select whole exome sample columns
- Move "Latest QC Complete Date" and "QC Complete Status" columns to the most right of the view (confirmed with Pav)

[#1631](https://app.zenhub.com/workspaces/smile-scrum-5fcfa28a4ccc1600165347b6/issues/gh/mskcc/smile-server/1631)

## Manual testing checklist

Check off items under the affected features below.

> [!NOTE]
> This PR only affects the UI and none of the functionalities below. I've tested to confirm that the tooltip appears and shows the expected text for each of the affected column header.

#### Searching & filtering

- [ ] Searching for a single term or multiple terms returns relevant results
- [ ] Searching in PHI mode returns results with PHI columns
- [ ] Filtering columns returns the results matching the filter
- [ ] [Samples page] Filtering by All, WES, and ACCESS/CMO-CH returns the expected results

#### Downloading

Confirm that these download actions return a TSV file with the expected columns and record count:

- [ ] Clicking "Download as TSV"
- [ ] Searching for some terms then clicking "Download as TSV"
- [ ] Filtering a column then clicking "Download as TSV"
- [ ] Search in PHI mode then clicking "Download as TSV"
- [ ] Clicking the dropdown download option <download option name>

#### AG Grid interactions

- [ ] [Samples page] Editing a cell updates the value as expected
- [ ] [Samples page] Pasting data into the grid works as expected
- [ ] Editing a cell in a non-editable row/column is prevented
- [ ] Column sorting in ascending and descending order works as expected
- [ ] PHI columns are visible only when PHI mode is enabled and search terms are present

#### Modals and popups

- [ ] [Samples page] Cell changes confirmation modal appears and works as expected
- [ ] [Requests page] Validation errors are displayed in the Record Validation modal
- [ ] Warning modals appear when expected and can be dismissed (e.g. PHI warning when first logged in)

#### Schema changes

Schema changes in [typeDefs.ts](./graphql-server/src/utils/typeDefs.ts) are reflected in the following:

- [ ] The queries defined in [operations.graphql](./graphql/operations.graphql)
- [ ] The GraphQL types that are auto-generated via `yarn run codegen`
- [ ] When applicable, ensure that newly added fields are searchable (i.e. added to the "searchable fields" list for a given query)

### DIFF
--- a/frontend/src/components/DataGrid.tsx
+++ b/frontend/src/components/DataGrid.tsx
@@ -40,7 +40,6 @@ const defaultColDef: ColDef = {
   resizable: true,
   editable: false,
   headerComponentParams: createCustomHeader(lockIcon),
-  valueFormatter: (params) => (params.value === "null" ? "" : params.value),
   tooltipValueGetter: (params: ITooltipParams) => getTooltipValue(params),
 };
 
--- a/frontend/src/components/DataGrid.tsx
+++ b/frontend/src/components/DataGrid.tsx
@@ -35,12 +35,21 @@ function getTooltipValue(params: ITooltipParams) {
   }
 }
 
+export function AgGridCustomTooltip(props: ITooltipParams) {
+  return (
+    <div className="ag-custom-tooltip">
+      <span>{props.value}</span>
+    </div>
+  );
+}
+
 const defaultColDef: ColDef = {
   sortable: true,
   resizable: true,
   editable: false,
   headerComponentParams: createCustomHeader(lockIcon),
   tooltipValueGetter: (params: ITooltipParams) => getTooltipValue(params),
+  tooltipComponent: AgGridCustomTooltip,
 };
 
 interface EditableGridProps {
--- a/frontend/src/index.scss
+++ b/frontend/src/index.scss
@@ -55,11 +55,6 @@ $standardMargin: 10px;
   cursor: not-allowed;
 }
 
-/* Prevent overlapping with the 🚫 icon */
-.ag-tooltip {
-  transform: translate(15px, -40px);
-}
-
 .costCenter-validation-error {
   color: #fa0400 !important;
 }
--- a/frontend/src/index.scss
+++ b/frontend/src/index.scss
@@ -150,3 +150,20 @@ $standardMargin: 10px;
 .popupTableHeight {
   height: calc(100vh - 270px);
 }
+
+// Enable the use of new lines (`\n`) in column headers
+// https://stackoverflow.com/a/57214867
+.ag-header-cell-label .ag-header-cell-text {
+  white-space: pre-wrap !important;
+}
+
+// Prevent column header icons from shrinking when the column width is too small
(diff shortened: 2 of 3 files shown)


frontend/src/components/DataGrid.tsx
```diff
@@ -40,7 +40,6 @@ const defaultColDef: ColDef = {
   resizable: true,
   editable: false,
   headerComponentParams: createCustomHeader(lockIcon),
-  valueFormatter: (params) => (params.value === "null" ? "" : params.value),
```
qu8n, line 43: This code hides any cell values that return a "null" string.

I've confirmed that none of the existing fields show "null" string values on the dashboard.

We don't want to hide this because:
1. When this issue occurs, this line of code doesn't prevent "null" string values from showing up in the downloaded file.
2. We can more quickly catch and resolve this error at the query-level.

frontend/src/index.scss
```diff
@@ -55,11 +55,6 @@ $standardMargin: 10px;
   cursor: not-allowed;
 }
 
-/* Prevent overlapping with the 🚫 icon */
-.ag-tooltip {
-  transform: translate(15px, -40px);
-}
-
```
qu8n, line 62: This is an artifact from an older iteration of the `"This column is read-only"` tooltip and is no longer relevant (no use of the 🚫 icon anywhere)

frontend/src/pages/samples/config.tsx
```diff
@@ -518,44 +530,60 @@ export const wesSampleColDefs: Array<ColDef<DashboardSample>> = [
...
-  {
-    field: "qcCompleteDate",
-    headerName: "Latest QC Complete Date",
-    valueFormatter: (params) => formatCellDate(params.value) ?? "",
-    ...getAgGridDateColFilterConfigs(),
+    headerName: "MAF Complete Status/Data Eligible for Sharing",
+    headerTooltip:
+      "Indicates whether MAF generation was successful. Valid for tumor samples only. For the MSK WES Repository cohort we only included samples with a MAF Complete Status of 'Pass' as well as QC Complete Result of 'Pass' or 'Warn'",
+    headerComponentParams: createCustomHeader(lockIcon + toolTipIcon),
+    width: 400,
```
ao508, line 570: Can we shrink the width of the tooltip box down? It's very wide.

ao508: Oh - Is it wide because of the new header name? Can we split the headerName to two lines and put "Data Eligible for Sharing" in parentheses in the second line?
```
MAF Complete Status
(Data Eligible for Sharing)
```

ao508 (changes requested):
Can we limit the width of the header tooltip boxes or is that dependent on the column that it belongs to?

ao508 (approved):

